PR #8, opened 2017-08-17
Added bower install into script-executor-plugin

rubij:


rubij:
Are we talking about execution-controller/src/main/resources/webapp/package.json ? It is the only place I found a reference to Gruntfile.js

jeromecomte:
I mean "execution-controller\src\main\resources\webapp\Gruntfile.js" which
should be executed with grunt after bower install

2017-08-18 9:43 GMT+02:00 rubij <notifications@github.com>:

PR #10, opened 2020-07-08
Rethink import export

david-stephan:
For review...


step-controller/step-controller-base-plugins/src/main/java/step/plugins/parametermanager/ParameterManagerPlugin.java
```diff
@@ -57,7 +57,7 @@
 	
 	private static final String PARAMETER_SCOPE_VALUE_DEFAULT = "default";
 	private static final String PARAMETERS_BY_SCOPE = "$parametersByScope";
-	private static final String entityName = "parameters";
+	public static final String entityName = "parameters";
```
jeromecomte, line 60: Minor: Java naming convention for constants

PR #12, opened 2021-03-24
ThreadGroup ramp packing option

fzawadiak:
Added option "Pack" requested b y Postfinance. In essence it will affect rampup in a way that users will be added not one by one but in predefined pack size. Default behavior would be pack 1.

Idea is to be easily do load test when we increase stepwise load every few minutes, so breakpoints would be more visible. Alternative approach could be to define multiple ThreadGroups but we consider it suboptimal and very timeconsuming for the tests that are planned.


david-stephan:
Thanks Filip for your contribution. After review of your change we see it as relevant and properly implemented, we merged it to the master, It will then be part of the next release.

PR #31, opened 2022-09-25
SED-1213-TS - Implements migration from measurements to buckets colle…

lucianbaciu:
…ction

### DIFF
--- a/step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesBucketingHandler.java
+++ b/step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesBucketingHandler.java
@@ -9,6 +9,7 @@ import step.core.timeseries.TimeSeriesIngestionPipeline;
 import step.plugins.measurements.Measurement;
 import step.plugins.measurements.MeasurementHandler;
 
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 
@@ -64,6 +65,22 @@ public class TimeSeriesBucketingHandler implements MeasurementHandler {
                 this.ingestionPipeline.ingestPoint(bucketAttributes, measurement.getBegin(), measurement.getValue());
             }
         });
+    }
+
+    /**
+     * This method will handle existing measurements, and will check if it is a gauge or normal measurement
+     * @param measurement
+     */
+    public void ingestExistingMeasurement(Measurement measurement) {
+        if (measurement == null) {
+            return;
+        }
+        measurement.remove("_id"); // because these measurements come with a generated id and can't be grouped into buckets.
+        if (measurement.getType().equals(THREAD_GROUP_MEASUREMENT_TYPE)) {
+            this.processGauges(List.of(measurement));
+        } else {
+            this.processMeasurement(measurement);
+        }
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesService.java
```diff
@@ -103,32 +100,46 @@ private TimeSeriesAggregationQuery mapToQuery(FetchBucketsRequest request) {
...
-            throw new ControllerServiceException("No measurement found matching this execution id");
         }
     }
+
+    @Operation(operationId = "checkTimeSeries", description = "Check if the time-series was created for a specific execution")
+    @Secured(right = "execution-read")
+    @GET
+    @Path("/execution/{executionId}/exists")
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
```
jeromecomte, line 135: Is the return type boolean compatible with @Produces(MediaType.APPLICATION_JSON)?

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesService.java
```diff
@@ -103,32 +100,46 @@ private TimeSeriesAggregationQuery mapToQuery(FetchBucketsRequest request) {
...
-                throw new ControllerServiceException("Time series already exist for this execution. Unable to rebuild it");
-            } else {
+        if (this.timeSeriesExists(executionId)) {
+            throw new ControllerServiceException("Time series already exist for this execution. Unable to rebuild it");
+        } else {
+            // we need to check if measurements exists
+            Equals measurementFilter = Filters.equals(MeasurementPlugin.ATTRIBUTE_EXECUTION_ID, executionId);
+            Measurement firstMeasurement = measurementCollection.find(measurementFilter,
+                    new SearchOrder(MeasurementPlugin.BEGIN, 1), 0, 1, 0).findFirst().orElse(null);
+            Measurement lastMeasurement = measurementCollection.find(measurementFilter,
```
jeromecomte, line 110: Why do we need to check the existence of the lastMeasurement? is the presence of the first one not enough?

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesBucketingHandler.java
```diff
@@ -66,6 +67,22 @@ public void processGauges(List<Measurement> measurements) {
...
 
+    /**
+     * This method will handle existing measurements, and will check if it is a gauge or normal measurement
+     * @param measurement
+     */
+    public void ingestExistingMeasurement(Measurement measurement) {
+        if (measurement == null) {
+            return;
+        }
+        measurement.remove("_id"); // because these measurements come with a generated id and can't be grouped into buckets.
```
jeromecomte, line 78: Removing the id could create side effects for the caller method. Would be better to ignore the Id durint the ingestion

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesBucketingHandler.java
```diff
@@ -66,6 +67,22 @@ public void processGauges(List<Measurement> measurements) {
...
+        if (measurement.getType().equals(THREAD_GROUP_MEASUREMENT_TYPE)) {
```
jeromecomte, line 79: Good, that you caught this special case. We might have to refactor the handling of gauges in the future in order not to have to treat them as special case...

jeromecomte:
@lucianbaciu Thanks for the changes and the handling of the special cases. See my comments...

PR #35, opened 2022-10-21
SED-1484-Make start/end mandatory for time series

lucianbaciu:


david-stephan (changes requested):
Hi @lucianbaciu , the change looks alright except that we should align the error handling. Please catch and rethrow a ControllerServiceException in the service getBuckets as done in the other services.

david-stephan (approved):

PR #40, opened 2022-12-06
SED-1606 Support Error Handlers for generic scripts

cl-exense:
This is already live and working at ADC as a customer-specific patch...

### DIFF
--- a/step-functions-plugins/step-functions-plugins-java/step-functions-plugins-java-def/src/main/java/step/plugins/java/GeneralScriptFunctionType.java
+++ b/step-functions-plugins/step-functions-plugins-java/step-functions-plugins-java-def/src/main/java/step/plugins/java/GeneralScriptFunctionType.java
@@ -57,7 +57,7 @@ public class GeneralScriptFunctionType extends AbstractScriptFunctionType<Genera
 	public Map<String, String> getHandlerProperties(GeneralScriptFunction function) {
 		String language = getScriptLanguage(function);
 		if (language != null) {
-			String errorHandler = configuration.getProperty("script." + language + "_errorhandler", null);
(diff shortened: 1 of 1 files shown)


step-functions-plugins/step-functions-plugins-java/step-functions-plugins-java-def/src/main/java/step/plugins/java/GeneralScriptFunctionType.java
```diff
@@ -49,4 +52,16 @@ public GeneralScriptFunction newFunction() {
 		function.getScriptLanguage().setValue("java");
 		return function;
 	}
+
+	@Override
+	public Map<String, String> getHandlerProperties(GeneralScriptFunction function) {
+		String language = getScriptLanguage(function);
+		if (language != null) {
+			String errorHandler = configuration.getProperty("script." + language + "_errorhandler", null);
```
jeromecomte, line 60: @cl-exense I would recommend to use the same namespace as the other configuration items of the AbstractScriptFunctionType: for instance: plugins.<language>.errorhandler

jeromecomte (changes requested):
See comment on line 60

step-functions-plugins/step-functions-plugins-java/step-functions-plugins-java-def/src/main/java/step/plugins/java/GeneralScriptFunctionType.java
cl-exense: OK, thanks -- done, now using plugins.`${language}`.errorhandler

jeromecomte (approved):

PR #41, opened 2022-12-07
Systemprop exec params

doriancransac:
Made the necessary changes for the two highlighted spots from the attached image to work.

image

I feel like the way the results are printed should be made configureable because it's very difficult to understand what's going on right now (even when there is a failure and the tree is printed). That way we may be able to achieve the same results without having to sign every NodeReport with a concrete toString() method.

What's your point of view on this?


doriancransac:
Sorry I should have probably picked a different target branch instead of master. Not sure if you can change it on your side..?

jeromecomte:
Created new pull request to branch SED-1581-support-environment-variables-in-step-j-unit-runner: https://github.com/exense/step/pull/42. Closing this one

PR #46, opened 2022-12-22
sed-1592 composite keywords plan reference

iegorov777:
This PR contains also the fix for SED-596 (deep copy for plans linked to composite)


david-stephan:
Hi @iegorov777 the change itself looks good, however few general remarks for your next PR
* please avoid reformatting the code (like switching from space to tabs for indentation)
* do not include commits from other PR (in this one for instance we also have a commit from SED-596).
This makes the review and the merge more complicated

david-stephan:
Did a cherry pick of the relevant commit to master for simplicity

PR #50, opened 2023-01-12
SED-726 improve api

jeromebrongniart:


step-functions/step-functions-package/src/main/java/step/functions/packages/handlers/JavaFunctionPackageDaemon.java
```diff
@@ -82,7 +83,10 @@ protected FunctionList getFunctions(DiscovererParameters parameters) {
 					GeneralScriptFunction function = new GeneralScriptFunction();
 					function.setAttributes(new HashMap<>());
 					function.getAttributes().put(AbstractOrganizableObject.NAME, functionName);
-					
+
+					function.getCallTimeout().setValue(annotation.timeout());
+					function.setDescription(annotation.description());
```
jeromecomte, line 88: @jeromebrongniart The default value of the field description of the Keyword annoation is an empty string. This would mean that the description field of the Function will be always set. Until now it was always null per default. How will be this handled especially in the FE? If I'm not wrong the FE is checking if the value is null or not

jeromecomte (commented):
- I couldn't find any JUnit test covering this new feature
- See question in code comment

step-functions/step-functions-package/src/main/java/step/functions/packages/handlers/JavaFunctionPackageDaemon.java
jeromebrongniart: As far as I can tell, this have no impact:
- The default value for the name is set the same way
- When the description is null, the field in the gui is empty. Same when having an empty scring

jeromecomte: @jeromebrongniart Just checked and found out that the FE is checking if the keyword has a description to display the tooltip (see l48 in https://github.com/exense/step-frontend/blob/60f45987b1fc46238527b432691189d5e6d231a8/projects/step-frontend/src/lib/angularjs/partials/dynamicforms/jsonEditor.html). If we keep your changes the FE should be adapted to check for non empty string

jeromebrongniart: @jeromecomte I rebuild (https://stepee-sed-726improve-api.exense.ch/#/root/functions?tenant=Common) and cannot see where the description is displayed. Note: the kw Open_Chrome have a description

I tested on the build server and I have the same prob. and cannot find where the description field is displayed

jeromecomte: @jeromebrongniart thanks for the test. the tooltip is indeed missing in step 20. I'll create a bug. Here were it should be displayed (tested on 3.18): 
image

jeromecomte (approved):
E2E tested. Approving and merging to master

PR #53, opened 2023-01-23
Sed 1581 support environment variables in step j unit runner

jeromecomte:


david-stephan:
@jeromecomte 99% looks good, but I have the filling that the tree is printed with all details only when the test pass. In the other caess, notifyFailure seems to call the printTree with node details to false: getExecutionTreeAsString(res) -> res.printTree(w, true); -> return printTree(writer, false, printAttachments);

jeromecomte:
@david-stephan Thanks a lot for the review. You're absolutely right. I'll change this

jeromecomte:
@david-stephan Report node details added in case of error.

david-stephan (approved):

PR #57, opened 2023-01-27
SED-1669 support simplified check syntax

david-stephan:


step-plans/step-plans-parser/src/main/java/step/repositories/parser/annotated/DefaultDescriptionStepParser.java
```diff
@@ -440,7 +440,7 @@ public static void echo(ParsingContext parsingContext, String expression) {
 		parsingContext.addArtefactToCurrentParent(result);
 	}
 	
-	@Step("Check (.*)$")
+	@Step(value="Check (Expression *=.*)$", priority=2)
```
jeromecomte, line 443: - This expression requires exactly one space between "Check" and "Expression" and might break some existing plans having more than one space.
- Keyvalue.g4 considers additional chars as space: [ \u00A0\t\r\n]. In order not to break existing expression I would also allow them as space

Would therefore recommend: Check[ \u00A0\t\r\n]+(Expression[ \u00A0\t\r\n]=.*)$

step-plans/step-plans-parser/src/test/java/step/repositories/parser/DefaultDescriptionStepParserTest.java
```diff
@@ -483,14 +484,30 @@ public void testIf() throws ParsingException {
...
 	@Test
 	public void testCheck() throws ParsingException {
 		List<AbstractStep> steps = new ArrayList<>();
-		steps.add(step("Check Expression= true \n"));
-
-		Check artefact = parseAndGetUniqueChild(steps, Check.class);
-		Assert.assertNotNull(artefact);
-		Assert.assertEquals("true",artefact.getExpression().getExpression());
+		steps.add(step("Check Expression= true \n "));
+		steps.add(step("Check Expression = key2.toString().equals('value2') \n "));
```
jeromecomte, line 492: Would add a test to cover the case with more than one space between Check and Expression: "Check   Expression=true"

jeromecomte (changes requested):
See line comments

step-plans/step-plans-parser/src/main/java/step/repositories/parser/annotated/DefaultDescriptionStepParser.java
david-stephan: Thanks, this indeed would have caused regressions, I updated as suggested.

step-plans/step-plans-parser/src/test/java/step/repositories/parser/DefaultDescriptionStepParserTest.java
david-stephan: Added more test to cover more spaces but also non breaking space, endlines ...

jeromecomte (approved):

PR #69, opened 2023-03-03
SED-1855 adding details when the function discoverer fail

jeromebrongniart:


david-stephan:
@jeromebrongniart : there is not much to say about the code. However what is the behaviour for the end user? If it is a technical/admin error, the details should be displayed in the logs rather than in the UI.

jeromebrongniart:
@david-stephan YOu are right: I changed the code to dump the error detail into a logger and only display a generic message back to the ui

david-stephan (approved):

PR #70, opened 2023-03-06
Sed 1306 ts   full interactive analysis based on raw measurements

lucianbaciu:


david-stephan (commented):
In addition to inline comments, please add new junit test for RAW measurements, OQL validation....

lucianbaciu:
@david-stephan regarding the tests, can you point me to a place where we test the endpoints, so I can follow?

david-stephan:
@lucianbaciu for junit related to endpoints, I have no location to share since in Step we usually avoid having much logic directly in the endpoint and test them e2e. Still it's doable just pass the context in a new constructor or give it a try with the Jersey test framework.

david-stephan (changes requested):
@lucianbaciu minor changes requested for the TimeSeriesHandler class

lucianbaciu:
I decided to not add Configuration object in the Handler, and pass the direct params instead

david-stephan (dismissed):
BE code change approved, some FE changes pending to be approved. We'll only merge to master  when all checks passed.

david-stephan (approved):

PR #71, opened 2023-03-09
Sed 1846 add system property

jeromebrongniart:
Fixing bug when trying to parse a SNAPSHOT file


david-stephan (commented):
Wouldn't it make sense to only set this sys property if it is unset?

PR #75, opened 2023-03-23
Sed 1846 add system property

jeromebrongniart:


david-stephan (changes requested):
I think we decided to only add the property if it is not set, letting the user overwirte it.

PR #77, opened 2023-03-23
SED-1889 Fixing NPE when no @Plans exist

jeromebrongniart:


david-stephan (approved):

jeromecomte (approved):
Approving this PR.
In the midterm one should ensure that StepJarParser and StepClassParser don't contain any duplicated code for the discover of plans as the bug fixed in this PR doesn't exist when running the plans using the Junit runner locally.

PR #82, opened 2023-04-04
Sed-1681 support composite in keyword package

iegorov777:
This branch includes SED-1679 (define composite as code) and applies the support of composite functions to JavaFunctionPackageDaemon.


david-stephan (changes requested):
see individual comments

david-stephan (dismissed):
For now accepting the code reformatting, but to avoided for next PRs, fixed dependencies versions and migration task Step version with last commits

david-stephan (approved):

PR #85, opened 2023-04-14
SED-1242-Added maven dependency management

lucianbaciu:
Added javax.validation in maven dependency management, so it can be easily imported (without a version) anywhere we define models and need to specify mandatory fields

### DIFF
--- a/step-parent/pom.xml
+++ b/step-parent/pom.xml
@@ -85,6 +85,16 @@
 		</repository>
 	</repositories>
 
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>javax.validation</groupId>
+                <artifactId>validation-api</artifactId>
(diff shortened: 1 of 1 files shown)


david-stephan (changes requested):
Step support java 11 and java 17, I would assume we should use the jakarta equivalent:
    <groupId>jakarta.validation</groupId>
    <artifactId>jakarta.validation-api</artifactId>

lucianbaciu:
Javax works in Java 11 and 17. Why do we need jakarta?

lucianbaciu:
@david-stephan nevermind. I see online that both works but it is recommended to use Jakarta for newer apps

david-stephan:
Jarkarta is indeed to be used, and we already have all required dependencies, declining this PR

PR #90, opened 2023-04-28
SED-1805-Added empty oql validation

lucianbaciu:


david-stephan (changes requested):
The indentation is incorrect, also the logical place for the newly added if block should be in the try block

david-stephan (approved):

PR #94, opened 2023-05-12
SED-2010 Integration with new Licensing Implementation

cl-exense:
Forgot to create this PR which contains a few changes required in the base. If we don't intend to use the token-seconds licensing (which is different from total runtime in that it only counts executions on actual agents), we can remove the TokenLifecycleInterceptor - related changes.


step-controller/step-controller-remote-client/src/main/java/step/client/functions/RemoteFunctionExecutionService.java
```diff
@@ -102,4 +103,14 @@ public Type getOwnerType() {
 		
 		return executeRequest(()->b.post(entity,genericType));
 	}
+
+	@Override
+	public void registerTokenLifecycleInterceptor(TokenLifecycleInterceptor interceptor) {
+		// FIXME: does this need an implementation?
```
david-stephan, line 109: @cl-exense this class is used by the step client, it will call the related REST service to reserve a token. The REST service  invokes the FunctionExecutionServiceImpl in the BE to actual get a token, so no implementation is required on the Remote class from my point of view.

david-stephan (changes requested):
@cl-exense I would keep the token interceptor part as we may have licensing on token usage per second. I think we're missing so test coverage for this implementation. I would cover also error cases when reserving / returning a tokens in the junits.

step-functions/step-functions-client-api/src/main/java/step/functions/execution/TokenLifecycleInterceptor.java
```diff
@@ -0,0 +1,7 @@
+package step.functions.execution;
+
+public interface TokenLifecycleInterceptor {
+    void onReturnTokenHandle(String tokenHandleId);
```
jeromecomte, line 4: Some listener might need the TokenWrapper (Quota by token type, etc). Instead of passing the Id I would therefore pass the token directly

step-functions/step-functions-client-api/src/main/java/step/functions/execution/TokenLifecycleInterceptor.java
```diff
@@ -0,0 +1,7 @@
...
+
+    void onGetTokenHandle(String tokenHandleId) throws Exception;
```
jeromecomte, line 6: Idem

step-functions/step-functions-client-api/src/main/java/step/functions/execution/TokenLifecycleInterceptor.java
jeromecomte, line 4: Some listener might need the TokenWrapper (Quota by token type, etc). Instead of passing the Id I would therefore pass the token directly

step-functions/step-functions-client/src/main/java/step/functions/execution/FunctionExecutionServiceImpl.java
```diff
@@ -102,27 +102,69 @@ public FunctionExecutionServiceImpl(GridClient gridClient, FunctionTypeRegistry
 	}
 
 	private static final Logger logger = LoggerFactory.getLogger(FunctionExecutionServiceImpl.class);
-	
+
+	private final List<TokenLifecycleInterceptor> tokenLifecycleInterceptors = new ArrayList<>();
```
jeromecomte, line 106: Using CopyOnWriteArrayList instead would remove the need for synchronization

step-functions/step-functions-client/src/main/java/step/functions/execution/FunctionExecutionServiceImpl.java
```diff
@@ -102,27 +102,69 @@ public FunctionExecutionServiceImpl(GridClient gridClient, FunctionTypeRegistry
 	}
 
 	private static final Logger logger = LoggerFactory.getLogger(FunctionExecutionServiceImpl.class);
-	
+
+	private final List<TokenLifecycleInterceptor> tokenLifecycleInterceptors = new ArrayList<>();
+
+	@Override
+	public void registerTokenLifecycleInterceptor(TokenLifecycleInterceptor interceptor) {
+		synchronized (tokenLifecycleInterceptors) {
```
jeromecomte, line 110: Not needed with CopyOnWriteArrayList

step-functions/step-functions-client-api/src/main/java/step/functions/execution/TokenLifecycleInterceptor.java
jeromecomte, line 6: Javadoc could be useful here. Especially to mention that an exception would block the token retrieval

step-functions/step-functions-client/src/main/java/step/functions/execution/FunctionExecutionServiceImpl.java
```diff
@@ -102,27 +102,69 @@ public FunctionExecutionServiceImpl(GridClient gridClient, FunctionTypeRegistry
...
 			throw new FunctionExecutionServiceException("Timeout after "+e.getCallTimeout()+"ms while reserving the agent token. You can increase the call timeout by setting 'grid.client.token.reserve.timeout.ms' in step.properties",e );
 		} catch (AgentSideException e) {
 			throw new FunctionExecutionServiceException("Unexpected error on the agent side while reserving the agent token: "+e.getMessage(),e);
 		} catch (AgentCommunicationException e) {
 			throw new FunctionExecutionServiceException("Communication error between the controller and the agent while reserving the agent token",e);
-		} 
+		}
+
+		try {
+			for (TokenLifecycleInterceptor interceptor : getTokenLifecycleInterceptors()) {
```
jeromecomte, line 142: Use forEach instead

step-functions/step-functions-client/src/main/java/step/functions/execution/FunctionExecutionServiceImpl.java
```diff
@@ -102,27 +102,69 @@ public FunctionExecutionServiceImpl(GridClient gridClient, FunctionTypeRegistry
...
+		synchronized (tokenLifecycleInterceptors) {
+			interceptors = new ArrayList<>(tokenLifecycleInterceptors);
+		}
+		return interceptors;
 	}
 
 	@Override
 	public void returnTokenHandle(String tokenHandleId) throws FunctionExecutionServiceException {
+		List<TokenLifecycleInterceptor> interceptors = getTokenLifecycleInterceptors();
+		interceptors.forEach(i -> i.onReturnTokenHandle(tokenHandleId));
```
jeromecomte, line 167: An error in the hooks would create a token leak. Shouldn't we catch exceptions in hooks and still return the token?

step-plans/step-plans-core/src/main/java/step/core/execution/ExecutionEngineRunner.java
```diff
@@ -69,23 +71,33 @@ protected PlanRunnerResult execute() {
...
-			if(plan == null) {
-				executionLifecycleManager.beforePlanImport();
-				
-				updateStatus(ExecutionStatus.IMPORTING);
-				ImportResult importResult = importPlan(executionContext);
-				
+			Plan plan = null;
+			List<ExecutionVeto> vetoes = executionLifecycleManager.getExecutionVetoes();
+			if (!vetoes.isEmpty()) {
+				logger.info("Execution was vetoed.");
```
jeromecomte, line 77: Would add the execution id to the log

step-functions/step-functions-client/src/main/java/step/functions/execution/FunctionExecutionServiceImpl.java
```diff
@@ -102,27 +102,69 @@ public FunctionExecutionServiceImpl(GridClient gridClient, FunctionTypeRegistry
...
+			}
+		} catch (Exception e) {
+			try {
+				returnTokenHandle(tokenWrapper.getID());
+			} catch (Exception ignored) {
+				logger.warn("Unexpected error while returning token handle, ignoring: ", e);
+			}
+			throw new FunctionExecutionServiceException("Error while retrieving agent token: " + e.getMessage(), e);
+		}
+		return tokenWrapper;
```
jeromecomte, line 153: Missing JUnit test coverage for the new hooks

jeromecomte (changes requested):

jeromecomte (changes requested):
JUnit test coverage should be extended to support the new hooks in FunctionExecutionService

step-functions/step-functions-client/src/main/java/step/functions/execution/FunctionExecutionServiceImpl.java
cl-exense: Does not work, because catching an Exception inside foreach is not possible.

step-functions/step-functions-client-api/src/main/java/step/functions/execution/TokenLifecycleInterceptor.java
cl-exense: Discussed, leaving as is for now.

step-functions/step-functions-client/src/main/java/step/functions/execution/FunctionExecutionServiceImpl.java
cl-exense: Done.

step-plans/step-plans-core/src/main/java/step/core/execution/ExecutionEngineRunner.java
cl-exense: Done

step-functions/step-functions-client-api/src/main/java/step/functions/execution/TokenLifecycleInterceptor.java
cl-exense: Documented in TokenLifecycleInterceptor class.

cl-exense:
@jeromecomte  @david-stephan All comments should be addressed now.

jeromecomte (approved):

cl-exense:
@david-stephan Do you mind approving this? Github is all red without you :)

PR #99, opened 2023-06-07
SED-2145 auto-number-of-threads-is-consumed-even-when-only-one-thread…

david-stephan:
…-is-required


step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/handlers/ForBlockHandler.java
```diff
@@ -152,7 +153,7 @@ public Consumer<DataPoolRow> createWorkItemConsumer(WorkerController<DataPoolRow
 						}
 					};
 				}
-			}, numberOfThreads);
+			}, numberOfThreads, OptionalInt.empty());
```
jeromecomte, line 156: Would recommend to add a new method consumeWork with the additional parameter in order to keep the signature and avoid this change here

step-plans/step-plans-core/src/main/java/step/threadpool/ThreadPool.java
```diff
@@ -135,7 +138,8 @@ public boolean isParallel() {
 	}
 
 	public <WORK_ITEM> void consumeWork(Iterator<WORK_ITEM> workItemIterator,
-			WorkerItemConsumerFactory<WORK_ITEM> workItemConsumerFactory, int numberOfThreads) {
+			WorkerItemConsumerFactory<WORK_ITEM> workItemConsumerFactory, int numberOfThreads,
+										OptionalInt expectedNumberOfThreads) {
```
jeromecomte, line 142: Would rename the parameter as follow:
- numberOfThreads => specifiedNumberOfThreads
- expectedNumberOfThreads => requiredNumberOfThreads

jeromecomte (approved):
Approving merge. See minor comments

step-plans/step-plans-core/src/main/java/step/threadpool/ThreadPool.java
david-stephan: done

PR #105, opened 2023-06-23
SED-2108-Finish discover API

lucianbaciu:


step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesHandler.java
```diff
@@ -34,7 +35,7 @@
 
 public class TimeSeriesHandler {
 
-	private static final int FIELDS_SAMPLING_LIMIT = 1_000;
+    private static final int FIELDS_SAMPLING_LIMIT = 1_000;
```
david-stephan, line 38: FIELDS_SAMPLING_LIMIT should be used as default value, but we should be able to set (optionally) a value from step.properties

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesService.java
```diff
@@ -98,5 +96,28 @@ public boolean timeSeriesIsBuilt(@PathParam("executionId") String executionId) {
     public Set<String> getMeasurementsAttributes(@QueryParam("filter") String oqlFilter) {
         return handler.getMeasurementsAttributes(oqlFilter);
     }
+	
+	@Secured(right = "execution-read")
```
david-stephan, line 100: indentation issues for @Secured and @Consumed for the 2 new services

david-stephan (changes requested):

david-stephan (approved):

PR #110, opened 2023-07-09
Sed 1535 improve running execution on behalf of

iegorov777:


step-controller/step-controller-server/src/main/java/step/core/deployment/AbstractStepServices.java
```diff
@@ -72,4 +74,16 @@ protected ObjectEnricher getObjectEnricher() {
 	protected ObjectFilter getObjectFilter() {
 		return objectHookRegistry.getObjectFilter(getSession());
 	}
+
+	protected AuthorizationManager<User, Session<User>> getAuthorizationManager(){
+		return getContext().get(AuthorizationManager.class);
```
jeromecomte, line 79: If the AuthorizationManager is required, it is recommended to use the methode getContext().require

step-controller/step-controller-server/src/main/java/step/core/execution/ExecutionServices.java
```diff
@@ -72,6 +72,7 @@ public void init() {
 	@Path("/start")
 	@Secured(right="plan-execute")
 	public String execute(ExecutionParameters executionParams) {
+		checkRightsOnBehalfOf("plan-execute", executionParams.getUserID());
```
jeromecomte, line 75: @iegorov777 @david-stephan The UserID is set by the REST client. If UserID is different from the authenticated user, shouldn't we apply additional checks? we could for instance add a right that allows users to run executions on behalf?

jeromecomte: Just saw that this is implemented on EE side. You can ignore the previous comment

jeromecomte (approved):
Approved with minor remark

PR #111, opened 2023-07-09
Sed 1675 allow assertion other than string with the assert control

iegorov777:


jeromecomte (changes requested):
Great PR. Clean code, good test coverage.
Following changes requested:
- Handling of expected value as Object (see comment below)
- Following additional operators should be supported "less than or equal", "greater than or equal", and "is null" ("is null" would return true if a value doesn't exist or is null)

iegorov777:
> Following additional operators should be supported "less than or equal", "greater than or equal", and "is null" ("is null" would return true if a value doesn't exist or is null)

To support null and missing values in "is null" operator  I was forced to change the logic in `AssertHandler` and remove the in-advance validation blocking nulls and undefined input values.

So there is a liitle change in behavior - now in case of missing value we fill the expected value in report node. You can see than in `AssertHandlerTest.testKeyDoesntExist()` test:
`assertEquals(".*al.*", child.getExpected())`

step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/handlers/AssertHandler.java
```diff
@@ -134,8 +109,134 @@ protected void execute_(AssertReportNode node, Assert artefact) {
...
+							+ actual.getClass().getSimpleName()
+							+ ") is not supported for operator " + operator;
+				}
+			}
+			return createFailedAssertResult(message);
+		}
+
+		if (!handler.isExpectedValueSupported(expectedValue)) {
+			String message;
+			if (expectedValue == null || (expectedValue instanceof String && ((String) expectedValue).isEmpty())) {
```
jeromecomte, line 196: @iegorov777 This path is not covered by JUnit tests. I would recommend to had a simple JUnit test that covers non supported expected values

step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/handlers/asserts/AbstractOperatorHandler.java
```diff
@@ -0,0 +1,59 @@
...
+    protected boolean isNull(Object value) {
+        return value == null;
+    }
+
+    protected int compareNumeric(Object actual, String expectedValueString) {
+        if (actual instanceof Integer) {
+            return ((Integer) actual).compareTo(Integer.valueOf(expectedValueString));
+        } else if (actual instanceof Long) {
+            return ((Long) actual).compareTo(Long.valueOf(expectedValueString));
+        } else if (actual instanceof Double) {
```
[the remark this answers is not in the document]
jeromecomte: @iegorov777 I've re-executed the tests with coverage. It seem that Long, Double and Float are still not covered by the JUnit tests of AssertHandlerTest

jeromecomte (changes requested):
@iegorov777 Thanks for the changes! I've implemented a JUnit test that tests dynamic expressions that returns Objects for expected values and I'm having following issue:

Following test fails:

		setupPassed();
		a = new Assert();
		a.setActual(new DynamicValue<String>("keyInt"));
		**a.setExpected(new DynamicValue<String>("777", ""));**
		a.setOperator(AssertOperator.EQUALS);

With following error "'keyInt' expected to be equal to '777' but was '777'"

As the dynmic expression is returning an int, I would expect it to pass.

Except for this, I've reported 2 minor remarks about test coverage.

step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/handlers/asserts/AbstractOperatorHandler.java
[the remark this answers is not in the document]
iegorov777: Double and Float were covered by separate tests (GreaterThanOperatorHandlerTest and LessThanOperatorHandlerTest), because it is hard to test them under AssertHandlerTest (when we use double/floats in json output data in this test, these values are automatically parsed as BigDecimals by javax.json, so we only can get BigDecimals in AssertHandler there).

FYI: In my last commit I've exteded GreaterThanOperatorHandlerTest and LessThanOperatorHandlerTest to test various types of expected values as well as types of actual values.

step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/handlers/AssertHandler.java
iegorov777: In last commit I've added some simple test cases for this path.

iegorov777: @jeromecomte 
> With following error "'keyInt' expected to be equal to '777' but was '777'"
This bug is also fixed under my last commit. Also I've changed the type of `expected` field from String to Object in AssertReportNode, because now we support various types of expected value.

step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/reports/AssertReportNode.java
```diff
@@ -28,7 +28,7 @@ public class AssertReportNode extends ReportNode {
 	
 	String key;
 	
-	String actual;
+	Object actual;
```
jeromecomte, line 31: @iegorov777 ReportNode belong to the model shared with the UI and REST services. Changing the type might impact the UI and Rest clients. These objects are also used in freemarker templates in the EE project step-reporting. I've checked in the UI and couldn't find usage of these fields. Rest clients might have to be adapted but if this change is required, we'll communicate it in the migration steps of the release

jeromecomte (dismissed):
Approving with warnings (see below)

step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/handlers/asserts/AbstractOperatorHandler.java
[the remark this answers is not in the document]
jeromecomte: @iegorov777 My bad. I haven't seen these 2 tests. Thanks for the additional tests

step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/reports/AssertReportNode.java
iegorov777: @jeromecomte 
To ensure the compatibily I can revert the type back to String here.

jeromecomte:
If that's not causing other issues, that might be safer.

step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/reports/AssertReportNode.java
iegorov777: I've changed the 'actual' value to String.

step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/handlers/AssertHandler.java
iegorov777: @jeromecomte 
I've re-checked the coverage. Now all these parts of code are covered either by AssertHandlerTest or by GreaterThanOperatorHandlerTest, EqualsOperatorHandlerTest etc.

jeromecomte (approved):

step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/reports/AssertReportNode.java
```diff
@@ -30,7 +30,7 @@ public class AssertReportNode extends ReportNode {
 	
 	String actual;
 	
-	String expected;
+	Object expected;
```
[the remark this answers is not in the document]
iegorov777: @jeromecomte this version in github file viewer seems to be outdated. In my last commit (10.08.2023) both `expected` and `actual` fields were changed to `String`.

david-stephan (approved):

PR #113, opened 2023-07-13
SED-2254 Plain text plan editor not working with PostgreSQL

david-stephan:


step-parent/pom.xml
```diff
@@ -56,7 +56,7 @@
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<!-- internal dependencies -->
 		<exense-commons.version>2.0.5</exense-commons.version>
-		<step-framework.version>2.0.8</step-framework.version>
+		<step-framework.version>2023.7.13-64afdec5285af4183e13721d</step-framework.version>
```
jeromecomte, line 59: @david-stephan shouldn't we wait for a release of the step-framework?
@david-stephan you've selected master a target branch. I This will end up in Step 22. I assume this is intended

jeromecomte (approved):

step-parent/pom.xml
david-stephan: @jeromecomte yes, If we want to fix this issue with Step 22.1 we'll need a new release of step-framework too. I just check there were no other changes since 2.0.8, so it's safe.

PR #117, opened 2023-07-24
SED-2074 adding performance unit tests during the nightly builds 

jeromebrongniart:


jeromebrongniart:
With this change the tests will run every nights

david-stephan (approved):

PR #118, opened 2023-08-02
SED-2222 add bulk selection to execution list

david-stephan:


step-controller/step-controller-server/src/main/java/step/core/execution/ExecutionServices.java
```diff
@@ -192,4 +202,41 @@ public Execution saveExecution(Execution execution) {
 	public void deleteExecution(@PathParam("id") String id) {
 		executionAccessor.remove(new ObjectId(id));
 	}
+
+	@Operation(description = "Start multiple executions according to the provided parameters.")
```
jeromecomte, line 206: Would name it "Restart"

jeromecomte (dismissed):

PR #121, opened 2023-08-14
Sed 2227 simplify reading keyword output syntax in plans

iegorov777:


step-plans/step-plans-base-artefacts/src/test/java/step/artefacts/handlers/UserFriendlyJsonObjectTest.java
david-stephan, line 1: @iegorov777 would it be possible to add some tests with groovy evaluation to check the syntax ouput.field for the different types?

david-stephan (changes requested):

david-stephan (approved):

PR #129, opened 2023-09-03
Sed 2391 maven plugin upload package functionality leak resources

iegorov777:


step-controller/step-controller-remote-client/src/main/java/step/client/resources/RemoteResourceManager.java
```diff
@@ -136,12 +137,32 @@ public Resource createResource(String resourceType, InputStream resourceStream,
 
 	@Override
 	public Resource createResource(String resourceType, boolean isDirectory, InputStream resourceStream, String resourceFileName,
-								   boolean checkForDuplicates, ObjectEnricher objectEnricher) throws IOException, SimilarResourceExistingException {
+								   boolean checkForDuplicates, ObjectEnricher objectEnricher) {
 		StreamDataBodyPart bodyPart = new StreamDataBodyPart("file", resourceStream, resourceFileName);
+
+		// !!! in fact, 'checkForDuplicates' parameter is ignored, because the list of found duplicated resources (with the same hash sums)
+		// is located in ResourceUploadResponse.similarResources, but we ignore this list here and just take the uploaded resource
```
jeromecomte, line 144: The local implementation of the ResourceManager (ResourceManagerImpl) throws a SimilarResourceExistingException when a similar resource already exists. In my opinion this remote implementation of the ResourceManager should behave the same. Any reason for not throwing a SimilarResourceExistingException in this case here?

step-controller/step-controller-remote-client/src/main/java/step/client/resources/RemoteResourceManager.java
```diff
@@ -136,12 +137,32 @@ public Resource createResource(String resourceType, InputStream resourceStream,
 
 	@Override
 	public Resource createResource(String resourceType, boolean isDirectory, InputStream resourceStream, String resourceFileName,
-								   boolean checkForDuplicates, ObjectEnricher objectEnricher) throws IOException, SimilarResourceExistingException {
+								   boolean checkForDuplicates, ObjectEnricher objectEnricher) {
 		StreamDataBodyPart bodyPart = new StreamDataBodyPart("file", resourceStream, resourceFileName);
+
+		// !!! in fact, 'checkForDuplicates' parameter is ignored, because the list of found duplicated resources (with the same hash sums)
+		// is located in ResourceUploadResponse.similarResources, but we ignore this list here and just take the uploaded resource
 		ResourceUploadResponse upload = upload(bodyPart, resourceType, isDirectory, checkForDuplicates);
```
jeromecomte, line 145: It the result of ResourceUploadResponse is always ignored here, couldn't we call the upload method wiht checkForDuplicates = false. This is much faster as it doesn't check for similar resouces

step-controller/step-controller-remote-client/src/main/java/step/client/resources/RemoteResourceManager.java
```diff
@@ -136,12 +137,32 @@ public Resource createResource(String resourceType, InputStream resourceStream,
...
+		// !!! in fact, 'checkForDuplicates' parameter is ignored, because the list of found duplicated resources (with the same hash sums)
+		// is located in ResourceUploadResponse.similarResources, but we ignore this list here and just take the uploaded resource
 		ResourceUploadResponse upload = upload(bodyPart, resourceType, isDirectory, checkForDuplicates);
 		return upload.getResource();
 	}
 
+	@Override
+	public Resource createOrReuseResource(String resourceType, InputStream resourceStream, String resourceFileName, ObjectEnricher objectEnricher) {
+		StreamDataBodyPart bodyPart = new StreamDataBodyPart("file", resourceStream, resourceFileName);
+		ResourceUploadResponse upload = upload(bodyPart, resourceType, false, true);
```
jeromecomte, line 152: It appears that this service looks up for similar resources in all projects. If a similar resource exists within another project this similar resource would be returned by this service, regardless if the project it belongs to is global or not. We should test this carefully and see if it doesn't cause any issue. In my opinion it would be cleaner to filter out similar resources from other non global projects in step.resources.ResourceManagerImpl#closeResourceContainer

jeromecomte (commented):
@iegorov777 PR reviewed. See open points

david-stephan (approved):

PR #130, opened 2023-09-07
SED-2095 Show correct results/status for vetoed executions and failed imports

cl-exense:


step-plans/step-plans-core/src/main/java/step/core/execution/ExecutionEngineRunner.java
```diff
@@ -108,8 +111,8 @@ protected PlanRunnerResult execute() {
 				
 				ReportNode rootReportNode = executionContext.getReport();
 				executionContext.setCurrentReportNode(rootReportNode);
-				persistReportNode(rootReportNode);
```
jeromecomte, line 111: Why this change?

step-plans/step-plans-core/src/main/java/step/core/execution/ExecutionEngineRunner.java
```diff
@@ -211,11 +213,27 @@ private void exportExecution(ExecutionContext context) {
...
+		executionContext.getReportNodeAccessor().save(contextReport);
+		// there must be at least one child element to the (context) report.
+		ReportNode rootReport = new ReportNode();
+		rootReport.setExecutionID(contextReport.getExecutionID());
+		rootReport.setParentID(contextReport.getId());
+		rootReport.setStatus(status);
+		try {
+			// This needs to be done like this -- it will work at runtime, but
+			// would create a compile-time cyclic dependency between modules if the dependencies were setup "correctly" in pom.xml
+			AbstractArtefact failure = (AbstractArtefact) Class.forName("step.artefacts.Failure").newInstance();
```
jeromecomte, line 232: Why do we need a report node for this? ReportNode are meant to report execution of artifacts. In the cas of an import error we have not plan to execute and thus no report node. Aren't we misusing the ReportNodes to have a result for the execution?

jeromecomte (commented):
See comments

step-plans/step-plans-core/src/main/java/step/core/execution/ExecutionEngineRunner.java
cl-exense: ... because it was used exactly once, so I inlined it. :)

step-plans/step-plans-core/src/main/java/step/core/execution/ExecutionEngineRunner.java
cl-exense: Because the report node is used to show the icon on the UI. Otherwise, you woult not have the < x > icon...

step-core/src/main/java/step/core/artefacts/reports/ReportNodeStatus.java
```diff
@@ -18,8 +18,14 @@
...
 package step.core.artefacts.reports;
 
+// ATTENTION: The order is important here, lower ordinals have higher "importance".
+// Keep RUNNING as the last ine in the definitions.
 public enum ReportNodeStatus {
-	
+
+	VETOED,
+
+	IMPORT_ERROR,
```
jeromecomte, line 27: These 2 new statuses are actually not related to report nodes and thus shouldn't be declared here. The issue is that we're using the root report node status as result of the execution. This is actually wrong and should be changed in the future. As this might be a larger refactoring I'm approving this PR for now and we should refactor this later

jeromecomte (approved):

PR #132, opened 2023-09-15
SED-2367-drilldown-from-one-analytics-view-to-the-related-execution

david-stephan:


step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesHandler.java
```diff
@@ -272,7 +272,9 @@ private TimeSeriesAggregationQuery mapToQuery(FetchBucketsRequest request, TimeS
                                 TimeSeriesFilterBuilder.buildFilter(request.getOqlFilter()),
                                 TimeSeriesFilterBuilder.buildFilter(request.getParams()))
                 ))
-                .withGroupDimensions(request.getGroupDimensions());
+                .withGroupDimensions(request.getGroupDimensions())
+                .withCollectAttributes(request.getCollectAttributeKeys(),
```
jeromecomte, line 276: Is request.getCollectAttributeKeys() always set? i.e. is it set to null or to an empty collection when attribute collection is not needed? If not needed we should ensure that the collection is null (which improves the performance of BucketBuilder.accumulateAttributes) and it would be in this case preferable not to call withCollectAttributes at all.

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/api/FetchBucketsRequest.java
```diff
@@ -11,6 +11,8 @@ public class FetchBucketsRequest {
     private Integer numberOfBuckets;
     private long intervalSize; // in ms
     private List<Integer> percentiles = Collections.emptyList();
+    private List<String> collectAttributeKeys;
```
jeromecomte, line 14: Isn't it a set?

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesHandler.java
david-stephan: Changed as requested; removed the check of "isEmpty" in BucketBuilder

jeromecomte (dismissed):

PR #147, opened 2023-10-30
SED-2539 improving error message

jeromebrongniart:


david-stephan (dismissed):

step-controller/step-controller-base-plugins/src/main/java/step/plugins/parametermanager/ParameterServices.java
```diff
@@ -114,7 +114,7 @@ private Parameter save(Parameter newParameter, Parameter sourceParameter) {
 		try {
 			newParameter = parameterManager.encryptParameterValueIfEncryptionManagerAvailable(newParameter);
 		} catch (EncryptionManagerException e) {
-			throw new ControllerServiceException("Error while encrypting parameter value");
+			throw new ControllerServiceException(e.getMessage());
```
jeromecomte, line 117: Would keep the message and use following constructor:
new ControllerServiceException("Error while encrypting parameter value", e)

jeromecomte (approved):

step-controller/step-controller-base-plugins/src/main/java/step/plugins/parametermanager/ParameterServices.java
jeromebrongniart: Done

PR #155, opened 2023-11-10
SED-2544-filemanager-automatic-cleanup

david-stephan:


step-functions/step-functions-docker-handler/src/main/java/step/grid/ProxyGridServices.java
```diff
@@ -24,7 +24,7 @@ public class ProxyGridServices {
     @Path("/file/{id}/{version}")
     public Response getFile(@PathParam("id") String id, @PathParam("version") String version) throws IOException, FileManagerException {
```
jeromecomte, line 25: @david-stephan Don't we want to expose the new cleanable parameter in this service? doesn't seem to be required right now but could be later.

jeromecomte (commented):
Looks good. Just having one question.

step-functions/step-functions-docker-handler/src/main/java/step/grid/ProxyGridServices.java
david-stephan: @jeromecomte : I added it as optional query parameter with default value true.

jeromecomte (approved):

PR #158, opened 2023-11-13
Sed 2516 yaml plan doesn't support for and for each control properly

iegorov777:


iegorov777:
These changed are already merged to SED-2518.

PR #159, opened 2023-11-15
SED-2557: maven plugin to upload automation package

iegorov777:


step-automation-packages/step-automation-packages-client/src/main/java/step/automation/packages/client/RemoteAutomationPackageClientImpl.java
```diff
@@ -0,0 +1,75 @@
...
+        return uploadPackage(automationPackageFile, multiPartEntity -> {
+            Invocation.Builder builder = requestBuilder("/rest/automation-packages");
+            return RemoteAutomationPackageClientImpl.this.executeRequest(() -> builder.post(multiPartEntity, String.class));
+        });
+    }
+
+    @Override
+    public String createOrUpdateAutomationPackage(File automationPackageFile) {
+        return uploadPackage(automationPackageFile, multiPartEntity -> {
+            Invocation.Builder builder = requestBuilder("/rest/automation-packages");
```
david-stephan, line 53: @iegorov777 The put service should have the package name in URI, should it be done as part of this CR?

iegorov777: @david-stephan
I thought about that, but there is a difficulty, because to pass the package name in URI we should either add this name to plugin configuration (and this name should be the same as in automation descriptor). 
Or we should extract the name from automation package file (unzip), which also doesn't make sense...

step-automation-packages/step-automation-packages-client/src/main/java/step/automation/packages/client/AutomationPackageClient.java
```diff
@@ -0,0 +1,31 @@
...
+ * along with STEP.  If not, see <http://www.gnu.org/licenses/>.
+ ******************************************************************************/
+package step.automation.packages.client;
+
+import java.io.Closeable;
+import java.io.File;
+
+public interface AutomationPackageClient extends Closeable {
+
+    String newAutomationPackage(File automationPackageFile);
```
jeromecomte, line 26: would call it "create"

jeromecomte (approved):

step-automation-packages/step-automation-packages-client/src/main/java/step/automation/packages/client/AutomationPackageClient.java
iegorov777: Renamed

PR #161, opened 2023-11-20
SED-2560: isolated execution of automation package

iegorov777:


step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/AutomationPackageManager.java
```diff
@@ -165,12 +174,16 @@ private void persistStagedEntities(List<ExecutiontTaskParameters> completeExecTa
         }
 
         for (ExecutiontTaskParameters execTasksParameter : completeExecTasksParameters) {
-            executionScheduler.addExecutionTask(execTasksParameter, false);
+            if (executionScheduler != null) {
+                executionScheduler.addExecutionTask(execTasksParameter, false);
```
jeromecomte, line 178: Would add a TODO to move this to a dedicated class as part of SED-2594

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/AutomationPackageServices.java
```diff
@@ -64,11 +70,31 @@ public void deleteAutomationPackage(@PathParam("name") String automationPackageN
...
+    // TODO: add new right to the permission matrix
+    @POST
+    @Consumes(MediaType.MULTIPART_FORM_DATA)
+    @Produces(MediaType.APPLICATION_JSON)
+    @Path("/execute")
+    @Secured(right = "automation-package-execute")
+    public List<String> executeAutomationPackage(@FormDataParam("file") InputStream automationPackageInputStream,
+                                                 @FormDataParam("file") FormDataContentDisposition fileDetail,
+                                                 @Context UriInfo uriInfo) throws Exception {
+        // TODO: execution parameters map?
```
jeromecomte, line 88: We should pass the ExecutionParameters (minus plan and repositoryObject) as in step.core.execution.ExecutionServices#execute

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/AutomationPackageServices.java
```diff
@@ -64,11 +70,31 @@ public void deleteAutomationPackage(@PathParam("name") String automationPackageN
...
+        // TODO: add filters for plans
```
jeromecomte, line 89: The filter for plans could be step.core.execution.model.ExecutionParameters#artefactFilter

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/execution/AutomationPackageExecutor.java
```diff
@@ -0,0 +1,112 @@
...
+                }
+                params.setCustomParameters(executionParameters);
+
+                String newExecutionId = this.scheduler.execute(params);
+                if (newExecutionId != null) {
+                    executions.add(newExecutionId);
+                }
+            }
+        } finally {
+            isolatedAutomationPackageRepository.clearContext(contextId.toString());
```
jeromecomte, line 95: scheduler.execute is asynchron. Shouldn't we wait for the executions to finish?

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/execution/IsolatedAutomationPackageRepository.java
```diff
@@ -0,0 +1,116 @@
...
+
+    @Override
+    public ArtefactInfo getArtefactInfo(Map<String, String> repositoryParameters) throws Exception {
+        AutomationPackage automationPackage = getAutomationPackageForContext(repositoryParameters);
+        if (automationPackage == null) {
+            return null;
+        }
+        ArtefactInfo info = new ArtefactInfo();
+        // TODO: type?
+        info.setType("automationPackage");
```
jeromecomte, line 51: The method getArtefactInf won't be used for the REST service /execute which will directly call importArtefact. It would be used only if we want to trigger an execution from the UI view #/root/repository. We might introduce a new type indeed

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/execution/IsolatedAutomationPackageRepository.java
```diff
@@ -0,0 +1,116 @@
...
+        AutomationPackageManager automationPackageManager = getContext(repositoryParameters);
+        AutomationPackage automationPackage = getAutomationPackageForContext(repositoryParameters);
+
+        String planId = repositoryParameters.get(RepositoryObjectReference.PLAN_ID);
+
+        Plan plan = automationPackageManager.getPackagePlans(automationPackage.getId().toString())
+                .stream()
+                .filter(p -> p.getId().toString().equals(planId)).findFirst().orElse(null);
+        if (plan == null) {
+            return null;
```
jeromecomte, line 77: Should return an error in ImportResult

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/execution/IsolatedAutomationPackageRepository.java
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        enrichPlan(context, plan);
+        context.getPlanAccessor().save(plan);
+
+        ImportResult result = new ImportResult();
+        result.setPlanId(plan.getId().toString());
+
+        plan.getFunctions().iterator().forEachRemaining(f -> (context.get(FunctionAccessor.class)).save(f));
+        automationPackageManager.getPackageFunctions(automationPackage.getId().toString()).forEach(f -> context.get(FunctionAccessor.class).save(f));
```
jeromecomte, line 87: Shouldn't we import resources (like functions and datasources)?

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/execution/IsolatedAutomationPackageRepository.java
```diff
@@ -0,0 +1,116 @@
...
+
+        result.setSuccessful(true);
+
+        return result;
+    }
+
+    @Override
+    public TestSetStatusOverview getTestSetStatusOverview(Map<String, String> repositoryParameters) throws Exception {
+        // TODO: ...
```
jeromecomte, line 96: Only used for UI view #/root/repository. This should maybe return the list of plans. To be discussed

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/execution/IsolatedAutomationPackageRepository.java
```diff
@@ -0,0 +1,116 @@
...
+        return null;
+    }
+
+    @Override
+    public void exportExecution(ExecutionContext context, Map<String, String> repositoryParameters) throws Exception {
+        // TODO: ...
```
jeromecomte, line 102: No need to implement

jeromecomte (commented):
@iegorov777 Looks already very good. See my comments

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/execution/AutomationPackageExecutor.java
iegorov777: @jeromecomte 
Yes, we should. 
But, as I understood, the REST Serivce still should return the response asynchonosly without wait for execution complete? And here I need to launch some background task to perform cleanup after execution finish (like in RemoteExecutionManager#waitForTermination). Am I right?

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/execution/IsolatedAutomationPackageRepository.java
```diff
@@ -69,22 +75,54 @@ public ImportResult importArtefact(ExecutionContext context, Map<String, String>
...
+            result.setErrors(List.of(planAccessor.getClass() + " is not layered"));
+            return result;
+        }
+
+        planAccessor.save(plan);
 
-        plan.getFunctions().iterator().forEachRemaining(f -> (context.get(FunctionAccessor.class)).save(f));
-        automationPackageManager.getPackageFunctions(automationPackage.getId().toString()).forEach(f -> context.get(FunctionAccessor.class).save(f));
+        FunctionAccessor functionAccessor = context.get(FunctionAccessor.class);
+        if(!(functionAccessor instanceof LayeredAccessor)){
```
jeromecomte, line 100: This check makes sense. Please introduce a generic method to check this to avoid duplicated code

jeromecomte (commented):
Looks good. Only thing is that I could find the place where we instantiate the "temporary" resource manager. We could discuss this today

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/execution/IsolatedAutomationPackageRepository.java
```diff
@@ -69,22 +75,54 @@ public ImportResult importArtefact(ExecutionContext context, Map<String, String>
...
+        automationPackageManager.getPackageFunctions(automationPackage.getId()).forEach(functionAccessor::save);
+
+        ResourceManager contextResourceManager = context.getResourceManager();
+        if (!(contextResourceManager instanceof LayeredResourceManager)) {
+            result.setErrors(List.of(contextResourceManager.getClass() + " is not layered"));
+            return result;
+        }
+
+        // import all resources from automation package to execution context by adding the layer to contextResourceManager
+        ((LayeredResourceManager) contextResourceManager).pushManager(automationPackageManager.getResourceManager());
```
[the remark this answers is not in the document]
iegorov777: @jeromecomte 
By current design for isolated execution we create a seperate instance of `AutomationPackageManager` initialized with in-memory accessors and local resource manager for each execution.

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/AutomationPackagePlugin.java
```diff
@@ -77,11 +81,22 @@ public void afterInitializeData(GlobalContext context) throws Exception {
...
 
     }
+
+    @Override
+    public void serverStop(GlobalContext context) {
+        super.serverStop(context);
+        try {
+            this.packageExecutor.shutdown();
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();  //set the flag back to true
```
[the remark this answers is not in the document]
iegorov777: @jeromecomte 
Handling interruption is always painful for me... 

The idea was to protect from cases, when the calling code continues the execution after interruption. Like in simple example below, where the `for` loop will still be continued after the JVM receives the interruption signal (in `MyRunnable` we don't pass the interruption to the calling code and `sleep` is executed, because the thread doesn't have the 'interrupted' flag set).

`class Scratch {
    public static void main(String[] args) throws InterruptedException {
        Runnable runnable = new MyRunnable();
        for (int i = 0; i < 1000; i++) {
            runnable.run();
            Thread.sleep(100);
        }
    }

    public static class MyRunnable implements Runnable {

        @Override
        public void run() {
            try {
                someInterruptableMethod();
            } catch (InterruptedException e) {
                // can be used to pass interruption to the calling code and break the loop on Thread.sleep()
                // Thread.currentThread().interrupt();
            }
        }

        public void someInterruptableMethod() throws InterruptedException {
            // some logic that can throw InterruptedException
        }
    }
}`

But you are right, in our case the `Thread.currentThread().interrupt()` looks unnecessary here. I'll remove it.

step-plans/step-plans-yaml-parser/src/main/java/step/plans/parser/yaml/YamlPlanReader.java
```diff
@@ -100,7 +100,8 @@ public YamlPlanReader(Supplier<ObjectId> idGenerator, Version currentVersion, bo
 				this.jsonSchema = readJsonSchema(jsonSchemaPath);
 			} else {
 				// resolve the json schema to use
-				List<String> jsonSchemasFromExtensions = CachedAnnotationScanner.getClassesWithAnnotation(YamlPlanReaderExtension.class).stream()
+				List<String> jsonSchemasFromExtensions = CachedAnnotationScanner.getClassesWithAnnotation("step.plans", YamlPlanReaderExtension.class, Thread.currentThread().getContextClassLoader())
```
jeromecomte, line 103: Use constant for "step.plans"

step-plans/step-plans-yaml-parser/src/main/java/step/plans/parser/yaml/deserializers/YamlRootArtefactDeserializer.java
```diff
@@ -107,7 +107,7 @@ protected List<YamlArtefactFieldDeserializationProcessor> prepareFieldProcessors
 
     protected List<YamlArtefactFieldDeserializationProcessor> getExtensions() {
         List<YamlArtefactFieldDeserializationProcessor> extensions = new ArrayList<>();
-        CachedAnnotationScanner.getClassesWithAnnotation(YamlPlanReaderExtension.class).stream()
+        CachedAnnotationScanner.getClassesWithAnnotation("step.plans", YamlPlanReaderExtension.class, Thread.currentThread().getContextClassLoader()).stream()
```
jeromecomte, line 110: Use constant

step-plans/step-plans-yaml-parser/src/main/java/step/plans/parser/yaml/schema/YamlPlanJsonSchemaGenerator.java
```diff
@@ -129,14 +129,14 @@ protected List<JsonSchemaFieldProcessor> prepareFieldProcessors() {
 
 	protected List<JsonSchemaFieldProcessor> getFieldExtensions() {
 		List<JsonSchemaFieldProcessor> extensions = new ArrayList<>();
-		CachedAnnotationScanner.getClassesWithAnnotation(YamlPlanReaderExtension.class).stream()
+		CachedAnnotationScanner.getClassesWithAnnotation("step.plans", YamlPlanReaderExtension.class, Thread.currentThread().getContextClassLoader()).stream()
```
jeromecomte, line 132: Use constant

step-plans/step-plans-yaml-parser/src/main/java/step/plans/parser/yaml/schema/YamlPlanJsonSchemaGenerator.java
```diff
@@ -129,14 +129,14 @@ protected List<JsonSchemaFieldProcessor> prepareFieldProcessors() {
...
 				.map(newInstanceAs(YamlPlanReaderExtender.class)).forEach(e -> extensions.addAll(e.getJsonSchemaFieldProcessingExtensions()));
 		return extensions;
 	}
 
 	protected List<JsonSchemaDefinitionCreator> getDefinitionsExtensions() {
 		List<JsonSchemaDefinitionCreator> extensions = new ArrayList<>();
-		CachedAnnotationScanner.getClassesWithAnnotation(YamlPlanReaderExtension.class).stream()
+		CachedAnnotationScanner.getClassesWithAnnotation("step.plans", YamlPlanReaderExtension.class, Thread.currentThread().getContextClassLoader()).stream()
```
jeromecomte, line 139: Use constant

step-plans/step-plans-yaml-parser/src/main/java/step/plans/parser/yaml/serializers/YamlRootArtefactSerializer.java
```diff
@@ -107,7 +107,7 @@ protected List<YamlArtefactFieldSerializationProcessor> prepareFieldProcessors()
 
     protected List<YamlArtefactFieldSerializationProcessor> getExtensions() {
         List<YamlArtefactFieldSerializationProcessor> extensions = new ArrayList<>();
-        CachedAnnotationScanner.getClassesWithAnnotation(YamlPlanReaderExtension.class).stream()
+        CachedAnnotationScanner.getClassesWithAnnotation("step.plans", YamlPlanReaderExtension.class, Thread.currentThread().getContextClassLoader()).stream()
```
jeromecomte, line 110: Use constant

step-repositories/step-repositories-artifact/src/main/java/step/repositories/artifact/StepJarParser.java
```diff
@@ -18,108 +18,61 @@
...
-    private final StepClassParser stepClassParser;
-    private AutomationPackageDescriptorReader automationPackageDescriptorReader;
+    private final AutomationPackageReader automationPackageReader;
     private final AutomationPackageKeywordsAttributesApplier automationPackagesKeywordAttributesApplier;
 
     public StepJarParser() {
         this(null);
     }
 
     public StepJarParser(ResourceManager resourceManager) {
```
jeromecomte, line 47: Do we still need the StepJarParser?

step-functions/step-functions-client-api/src/main/java/step/functions/execution/FunctionExecutionService.java
```diff
@@ -39,6 +40,12 @@ public interface FunctionExecutionService {
 	
 	void returnTokenHandle(String tokenHandleId) throws FunctionExecutionServiceException;
 
-	<IN,OUT> Output<OUT> callFunction(String tokenHandleId, Function function, FunctionInput<IN> functionInput, Class<OUT> outputClass);
+	/**
+	 * Call function from while running the execution
+	 */
+	default <IN,OUT> Output<OUT> callFunction(String tokenHandleId, Function function, FunctionInput<IN> functionInput, Class<OUT> outputClass, AbstractStepContext executionContext){
```
jeromecomte, line 46: **To clarify:** Insteadof changing the signature of this method I'm wondering if it wouldn't be easier and cleaner to create a copy of the FunctionExecutionServiceImpl for each execution and store it to the context like here: step.engine.plugins.FunctionPlugin#initializeExecutionEngineContext

jeromecomte (changes requested):
Before merging this PR the following points should be discussed / clarified:
- Ensure that file resolver cache also works for automation package executions
- Check if it would be possible to create a copy of FunctionExecutionService with the ExecutionContext for each execution of an automation package (instead of changing the signature of FunctionExecutionService.callFunction)
- Check if StepJarParser is still used

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/AutomationPackageManager.java
```diff
@@ -172,47 +230,51 @@ private void persistStagedEntities(List<ExecutiontTaskParameters> completeExecTa
...
                 enricher.accept(e);
             }
         }
     }
 
-    protected List<Plan> preparePlansStaging(AutomationPackage newPackage, AutomationPackageContent packageContent, AutomationPackage oldPackage, ObjectEnricher enricher) {
+    protected List<Plan> preparePlansStaging(AutomationPackageContent packageContent, AutomationPackage oldPackage, ObjectEnricher enricher) {
         List<Plan> plans = packageContent.getPlans();
-        fillEntities(plans, newPackage, oldPackage != null ? getPackagePlans(oldPackage) : new ArrayList<>(), enricher);
+        fillEntities(plans, oldPackage != null ? getPackagePlans(oldPackage.getId()) : new ArrayList<>(), enricher);
```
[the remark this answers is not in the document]
iegorov777: @jeromecomte 
It seems we are not modifying old stored objects here. 

`getPackagePlans()` loads old plans via accessor (by the way, the local accessor should clone the original object because of `byPassCloning` property in `InMemoryCollection`). But these objects are not updated in `fillEntities()`. Those are only used to prepare the `nameToIdMap` and set the original ids to `plans` taken from `packageContent`,

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/AutomationPackageManager.java
```diff
@@ -263,88 +325,136 @@ protected AutomationPackage createNewInstance(String fileName, AutomationPackage
...
-
-    protected List<ExecutiontTaskParameters> deleteTasks(AutomationPackage automationPackage) {
-        List<ExecutiontTaskParameters> tasks = getPackageTasks(automationPackage);
-        tasks.forEach(task -> {
+    protected List<ExecutiontTaskParameters> deleteSchedules(AutomationPackage automationPackage) {
+        List<ExecutiontTaskParameters> schedules = getPackageSchedules(automationPackage.getId());
+        schedules.forEach(schedule -> {
             try {
-                executionScheduler.removeExecutionTask(task.getId().toString());
+                if (executionScheduler != null) {
```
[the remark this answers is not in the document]
iegorov777: @jeromecomte 
I agree. I think we could do this under the ticket, where we'll try to avoid the `executionScheduler` in `AutomationPackageManager`.

step-functions/step-functions-client-api/src/main/java/step/functions/execution/FunctionExecutionService.java
iegorov777: @jeromecomte 
I'll think how to improve this. The main reason why I changed the signature is that the `registerFile()` is finally called from the functionType object (like `JMeterFunctionType`). 

But each FunctionType in system is registered in `functionTypeRegistry` via `@Plugin` (for instance, `JMeterPlugin#serverStart()`).

So `functionType` are singletones, and it is hard to bind them to current execution context without adding the context to `AbstractFunctionType#getHandlerProperties()`. That is, the signature of `AbstractFunctionType#getHandlerProperties()` will be changed anyway...

step-repositories/step-repositories-artifact/src/main/java/step/repositories/artifact/StepJarParser.java
iegorov777: @jeromecomte 
`StepJarParser` is still used in `AbstractArtifactRepository`. When we are working with function package (for instance execute the function package via maven plugin), the `StepJarParser#getPlansForJar` is called.

I think, we can completely remove the `StepJarParser` when we stop supporting function packages and rewrite the `AbstractArtifactRepository`.

Or you want to move the remaining code from `StepJarParser` somewhere to the automation package module right now and delete the `StepJarParser` class?

step-automation-packages/step-automation-packages-support/src/main/java/step/automation/packages/AutomationPackageArchive.java
```diff
@@ -36,14 +36,20 @@ public class AutomationPackageArchive implements Closeable {
     public static final List<String> METADATA_FILES = List.of("automation-package.yml", "automation-package.yaml");
 
     private final ClassLoader classLoader;
+    private final File originalFile;
+    private final String originalFileName;
```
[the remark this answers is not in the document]
iegorov777: @jeromecomte 
It is different only in case of temporary files stored from incoming `InputStream`. If we avoid temp files (with changed names), we can remove the `originalFileName`.

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/AutomationPackageManager.java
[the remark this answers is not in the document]
jeromecomte: @iegorov777 Agree. I've added it to the ticket

step-core/src/main/java/step/resources/LayeredResourceManager.java
```diff
@@ -0,0 +1,177 @@
...
+    @Override
+    public void deleteResource(String resourceId) {
+        for (ResourceManager resourceManager : resourceManagers) {
+            resourceManager.deleteResource(resourceId);
+        }
+    }
+
+    @Override
+    public List<Resource> findManyByCriteria(Map<String, String> criteria) {
+        return null;
```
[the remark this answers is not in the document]
iegorov777: @jeromecomte 
Thanks. I've implemented this method.

step-functions/step-functions-client-api/src/main/java/step/functions/type/AbstractFunctionType.java
```diff
@@ -125,28 +132,58 @@ public T copyFunction(T function) throws FunctionTypeException {
...
+							ResourceManager unwrappedManager = unwrapResourceManager((LayeredResourceManager) executionContextResourceManager);
+							if (unwrappedManager == getResourceManager(null)) {
+								file = fileResolverCache.get(filepath);
+							} else {
+								resolvedFromExecutionContext = true;
+								file = new FileResolver(executionContextResourceManager).resolve(filepath);
+							}
+						} else {
+							resolvedFromExecutionContext = true;
+							file = new FileResolver(executionContextResourceManager).resolve(filepath);
```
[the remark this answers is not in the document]
iegorov777: @jeromecomte 

This sounds good. But for me it is not clear how to instantiate the `FileResolverCache` inside the `ExecutionContext`.
As I understood, in `AbstractFunctionType` we create an individual cache for each function type, and we apply the `FunctionTypeConfiguration` to `CacheBuilder` to build the cache properly.

In case we store the cache within the  `ExecutionContext` should we create some map `FunctionType` -> `FileResolverCache` in `ExecutionContext`?

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/execution/IsolatedAutomationPackageRepository.java
iegorov777: @jeromecomte 
Now it is solved via 
`((LayeredResourceManager) contextResourceManager).pushManager(automationPackageManager.getResourceManager())`

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/execution/AutomationPackageExecutor.java
```diff
@@ -0,0 +1,175 @@
...
+        }
+    }
+
+    protected void cleanupContextAfterAllExecutions(ObjectId contextId, AutomationPackageManager packageManager, List<String> executions, String fileName) {
+        // wait for all executions to be finished
+        delayedCleanupExecutor.execute(() -> {
+            if (waitForAllExecutionEnded(executions)) return;
+
+            // remove stored resources
+            packageManager.getResourceManager().cleanup();
```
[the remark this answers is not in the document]
iegorov777: @jeromecomte 
Good idea

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/AutomationPackageReader.java
```diff
@@ -77,10 +219,144 @@ public void fillAutomationPackageWithImportedFragments(AutomationPackageContent
...
+
+    private List<StepClassParserResult> getPlanFromPlansAnnotation(Class<?> klass, File artifact) {
+        final List<StepClassParserResult> result = new ArrayList<>();
+
+        Plans plans = klass.getAnnotation(Plans.class);
+        if (plans != null) {
+            for (String file : plans.value()) {
+                StepClassParserResult parserResult = null;
+                try {
+                    URL url = null;
```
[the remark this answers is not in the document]
iegorov777: @jeromecomte 
I have the same question. This code was moved 'as is' from `StepJarParser`, where the logic was the same:
* the jar file is already scanned by the classloader within `extractAnnotatedPlans()`
* but after that we try to load plan text or yaml plans listed in `@Plans` with new `JarURLConnection`

```
                    if (file.startsWith("/")) {
                        url = new URL("jar:file:" + artifact.getAbsolutePath() + "!" + file);
                    } else {
                        url = new URL("jar:file:" + artifact.getAbsolutePath() + "!/" +
                                klass.getPackageName().replace(".", "/") + "/" + file);
                    }

                    JarURLConnection jarURLConnection = (JarURLConnection) url.openConnection();                 
```
I believe there was some reason of doing that

step-functions/step-functions-client-api/src/main/java/step/functions/type/AbstractFunctionType.java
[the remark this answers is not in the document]
iegorov777: @jeromecomte 
Finally I've added the `fileResolverCache` field to the `AbstractStepContext` (configured with default concurrency level and maximum size).

step-functions-plugins/step-functions-plugins-composite/step-functions-plugins-local-composite/src/main/java/step/plugins/functions/types/LocalCompositeFunctionPlugin.java
```diff
@@ -60,18 +63,27 @@ public void initializeExecutionEngineContext(AbstractExecutionEngineContext pare
 
 	public void saveLocalFunctions() {
 		Set<Method> methods = CachedAnnotationScanner.getMethodsWithAnnotation(Keyword.class);
+		List<Function> functions = getLocalCompositeFunctions(methods);
```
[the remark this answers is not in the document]
iegorov777: @iegorov777 
To run the automation package it looks redundant, but for local runs via `step.junit.runner.Step` it iseems to be required to store all composite functions via `LocalCompositeFunctionPlugin` to execute plan.

PR #162, opened 2023-11-30
Sed 2530 ability to have multiple analytics dashboards

lucianbaciu:


step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/dashboards/model/ChartFilterItem.java
```diff
@@ -0,0 +1,104 @@
+package step.plugins.timeseries.dashboards.model;
+
+import jakarta.validation.constraints.NotNull;
+
+import java.util.List;
+
+public class ChartFilterItem {
```
david-stephan, line 7: For all classes with Chart in their name, are they restricted to chart or apply to dashblet. In the later case, they should be renamed

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/dashboards/model/ChartFilterItem.java
```diff
@@ -0,0 +1,104 @@
...
+	
+	private String label; // optional. should be used for filter options
```
david-stephan, line 9: Seems this class is having optional fields depending on the filter type, shouldn't we have a class hierarchy here?

step-core/src/main/java/step/core/entities/EntityManager.java
```diff
@@ -46,6 +46,7 @@ public class EntityManager  {
 	public final static String recursive = "recursive";
 	public final static String measurements = "measurements";
     public final static String metricTypes = "metricTypes";
```
david-stephan, line 48: indentattion issue

david-stephan (changes requested):

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/dashboards/model/ChartFilterItem.java
lucianbaciu: @david-stephan they are used in dashlets, but also in the dashboard (the global filters). maybe call them TimeSeriesFilterItem?

david-stephan: @lucianbaciu ultimetaley global filters are used by dashlets too, but TimeSeriesFilterItem is fine with me.

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/dashboards/model/ChartFilterItem.java
lucianbaciu: @david-stephan all the attributes are together now. We can split into multiple classes based on the type, but I'm not sure it is worth it. This way the processing is very straight in both FE and BE.

david-stephan (changes requested):
@lucianbaciu See the individual comment, I will directly commit changes for all of them except for the cleanup of "createTestDashboard" and extension of the junit coverage for dashboard. Please take care of that.

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesControllerPlugin.java
```diff
@@ -25,113 +31,244 @@
...
+	}
+
+	private void createLegacyDashboard(DashboardAccessor dashboardAccessor) {
+		boolean legacyDashboardExists = dashboardAccessor.stream().anyMatch(d -> d.getMetadata().get("isLegacy") != null && d.getMetadata().get("isLegacy").equals(true));
+		if (!legacyDashboardExists) {
+			DashboardView dashboard = new DashboardView();
+			dashboard.setName("Performance Dashboard");
+			dashboard.getMetadata().put("isLegacy", true);
+			dashboard.getMetadata().put("link", "analytics");
+			dashboardAccessor.save(dashboard);
```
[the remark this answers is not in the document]
lucianbaciu: @david-stephan I don't see any edit regarding this. do you want me to do it?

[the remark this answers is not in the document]
david-stephan: @lucianbaciu this is handle in the new EE plugin.

lucianbaciu:
@david-stephan regarding test coverage, the dashboard service provides nothing more than the CRUD + your method accessor method findLegacyDashboards() which may be tested because has custom logic.
- Do we need to test the CRUD, or better make tests in the framework when we move the models there?
- If yes, how should I test the service? should I follow TimeSeriesExecutionPluginTest and create the entire contexts/plugins etc? I don't know how to inject the service otherwise.

david-stephan:
@lucianbaciu there is no point adding unit test on the services, there is no new logic here. You're adding many classes as part of this PR with new accessor and model, make sure to have coverage for this code and definitely cleanup the unused function in the plugin class

PR #165, opened 2023-12-06
SED-2643 update logback

jeromebrongniart:


step-plans/step-plans-base-artefacts/pom.xml
```diff
@@ -96,7 +96,7 @@
 		</dependency>
 		<dependency>
 			<groupId>org.slf4j</groupId>
-			<artifactId>slf4j-api</artifactId>
+			<artifactId>log4j-over-slf4j</artifactId>
```
david-stephan, line 99: @jeromebrongniart why are we replacing the sl4j-api by log4j over slf4j here?

jeromebrongniart: @david-stephan org.apache.poi rely on log4j, This is the proper dependency if we want the logs.
In addition, slf4j-api is already imported by step-parent so unneeded here

PR #167, opened 2023-12-13
SED-1896-support-locking-plans-to-prevent-accidental-changes

david-stephan:


step-controller/step-controller-server/src/main/java/step/controller/services/entities/AbstractEntityServices.java
```diff
@@ -197,4 +198,36 @@ public History(String id, long updateTime) {
     public T restoreVersion(@PathParam("id") String id, @PathParam("versionId") String versionId) {
         return accessor.restoreVersion(new ObjectId(id), new ObjectId(versionId));
     }
+
+    @Operation(operationId = "is{Entity}Locked", description = "Get entity locking state")
+    @GET
+    @Path("/{id}/locked")
+    @Produces(MediaType.APPLICATION_JSON)
```
jeromecomte, line 205: Not sure if the boolean result of the method matches with this. This might create errors in client generators

step-controller/step-controller-server/src/main/java/step/controller/services/entities/AbstractEntityServices.java
```diff
@@ -197,4 +198,36 @@ public History(String id, long updateTime) {
...
+    @Secured(right = "{entity}-read")
+    public boolean isLocked(@PathParam("id") String id) {
+        T t = getEntity(id);
+        Boolean locked = t.getCustomField(CUSTOM_FIELD_LOCKED, Boolean.class);
+        return (locked != null && locked);
+    }
+
+    @Operation(operationId = "lock{Entity}", description = "Lock this entity")
+    @POST
+    @Produces(MediaType.APPLICATION_JSON)
```
jeromecomte, line 215: Not sure if it applies to void method

step-controller/step-controller-server/src/main/java/step/controller/services/entities/AbstractEntityServices.java
```diff
@@ -197,4 +198,36 @@ public History(String id, long updateTime) {
...
+    @Consumes(MediaType.APPLICATION_JSON)
```
jeromecomte, line 216: Not sure if the boolean payload matches with this. This might create errors in client generators

jeromecomte (commented):
Do I understand correctly that the locking fully relies on the client i.e. on the fact that the client asks for the locked status before editing and that no BE locking is implemented in the security filters?

step-controller/step-controller-server/src/main/java/step/controller/services/entities/AbstractEntityServices.java
david-stephan: this is valid for boolean and work with the FE client genetor

jeromecomte (approved):

PR #168, opened 2023-12-17
Sed 2627 support alerting rules in automation packages

iegorov777:


step-automation-packages/step-automation-packages-support/src/main/java/step/automation/packages/AutomationPackageAlertingRuleManager.java
```diff
@@ -0,0 +1,31 @@
...
+ ******************************************************************************/
+package step.automation.packages;
+
+import org.bson.types.ObjectId;
+import step.automation.packages.model.AutomationPackageAlertingRule;
+import step.core.objectenricher.ObjectEnricher;
+
+import java.util.List;
+
+public interface AutomationPackageAlertingRuleManager {
```
jeromecomte, line 27: This class seems to be a generic interface of the AlertingRuleManager. I would simply call it AlertingRuleManager

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/AutomationPackageManager.java
```diff
@@ -160,6 +167,16 @@ private void deleteAutomationPackageEntities(AutomationPackage automationPackage
         deletePlans(automationPackage);
         deleteSchedules(automationPackage);
         deleteResources(automationPackage);
+        deleteAlertingRules(automationPackage);
```
jeromecomte, line 170: In a second iteration we should do all this in a pluggable way i.e. the AutomationPackageManager should expose the hooks to which the different managers should register to

step-automation-packages/step-automation-packages-schema/src/main/java/step/automation/packages/yaml/schema/YamlAutomationPackageSchemaGenerator.java
```diff
@@ -60,12 +64,15 @@ public JsonNode generateJsonSchema() throws JsonSchemaPreparationException {
         JsonObjectBuilder allDefs = keywordSchemaGenerator.createKeywordDefs()
                 .addAll(planSchemaGenerator.createDefs())
                 .addAll(scheduleSchemaGenerator.createScheduleDefs());
+        if (supportAlertingRules) {
```
jeromecomte, line 67: OK for a first implementation. In a second iteration this should be implemented in a pluggable way

step-automation-packages/step-automation-packages-schema/src/main/java/step/automation/packages/yaml/schema/YamlAlertingRuleSchemaGenerator.java
```diff
@@ -0,0 +1,217 @@
...
+import step.plugins.alerting.event.AlertingEventInterface;
+import step.plugins.alerting.rule.condition.AlertingRuleAction;
+import step.plugins.alerting.rule.condition.AlertingRuleCondition;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+import static step.core.scanner.Classes.newInstanceAs;
+
+public class YamlAlertingRuleSchemaGenerator {
```
jeromecomte, line 43: Do we need a custom schema generator for each model class? can't we do this in a generic way for "simple" model classes like AutomationPackageAlertingRule?

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/deserialization/YamlAlertingRuleActionDeserializer.java
```diff
@@ -0,0 +1,65 @@
...
+import step.core.yaml.deserializers.StepYamlDeserializer;
+import step.core.yaml.deserializers.StepYamlDeserializerAddOn;
+import step.core.yaml.deserializers.YamlFieldDeserializationProcessor;
+import step.plugins.alerting.rule.condition.AlertingRuleAction;
+
+import java.io.IOException;
+import java.util.List;
+
+@StepYamlDeserializerAddOn(targetClasses = {AlertingRuleAction.class})
+public class YamlAlertingRuleActionDeserializer extends StepYamlDeserializer<AlertingRuleAction> {
```
jeromecomte, line 36: Can't we achieve this with jackson annotations?

step-core-model/src/main/java/step/plugins/alerting/rule/condition/AlertingRuleAction.java
```diff
@@ -0,0 +1,37 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with STEP.  If not, see <http://www.gnu.org/licenses/>.
+ ******************************************************************************/
+package step.plugins.alerting.rule.condition;
+
+public abstract class AlertingRuleAction {
```
jeromecomte, line 21: In the longterm all these classes should remain part of EE and registered to the package manager via plugin

jeromecomte (commented):
@iegorov777 Let's discuss on Tuesday if we want to merge this as a first iteration or try to implement the remarks related to pluggability and separation between OS and EE first

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/deserialization/YamlAlertingRuleActionDeserializer.java
iegorov777: @jeromecomte 
I agree that jackson annotations would be preferrable, but I wasn't able to perform this deserialization with pure jackson mostly because of existing `@JsonTypeInfo` annotations over target classes.
For yaml format we usually want to specify the type on the top level (to support auto-suggestions via json schema):

```
MyClass:
   fieldA: valueA
   fieldB: valueB
```
But I did't fild a way to apply this format with only jackson annotations - jackson forces to use the `class` field from `@JsonTypeInfo` even if I configure the yaml object mapper with another deserilizer.
Maybe it is possible, but I had not enough time to ivestigate that deeply.

step-automation-packages/step-automation-packages-schema/src/main/java/step/automation/packages/yaml/schema/YamlAlertingRuleSchemaGenerator.java
iegorov777: @jeromecomte 
I also would like to apply some generic way to generate a json schema for any model class, but we have to solve the issue with inherited classes and `@JsonTypeInfo` (from my previous comment).

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/AutomationPackageManager.java
iegorov777: @jeromecomte 
Fully agree. I think the pluggable implementation should cover both `AutomationPackageManager` and the `YamlAutomationPackageSchemaGenerator`.

step-core-model/src/main/java/step/plugins/alerting/rule/condition/AlertingRuleAction.java
iegorov777: @jeromecomte 
I agree. I moved some abstract classes and interfaces from EE to OS just because the implementations of automation package and automation package manager are now in OS, so we need to have some common interfaces these (to be used in `YamlAutomationPackageSchemaGenerator` and for deserialziation).

jeromecomte (approved):

PR #184, opened 2024-01-14
Sed 2705 unable to define alerting rules in yaml

iegorov777:


step-core/src/main/java/step/core/yaml/deserializers/CustomYamlFormat.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Retention(RUNTIME)
+@Target(ElementType.TYPE)
+public @interface CustomYamlFormat {
```
jeromecomte, line 29: Add javadoc to this class to explain its purpose

jeromecomte (changes requested):

PR #188, opened 2024-01-16
SED-2701 temporary-ap-resources-not-cleanup-properly-after-ap-update

david-stephan:


step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/AutomationPackageManager.java
```diff
@@ -318,11 +314,22 @@ protected Staging createStaging(){
...
                                          ObjectEnricher objectEnricher) {
+        List<Resource> stagingResources = staging.resourceManager.findManyByCriteria(null);
+        try {
+            for (Resource resource: stagingResources) {
+                resourceManager.copyResource(resource, staging.resourceManager);
+            }
+        } catch (IOException | SimilarResourceExistingException | InvalidResourceFormatException e) {
+            throw new AutomationPackageManagerException("Unable to persist a resource in automation package", e);
+        } finally {
+            staging.resourceManager.cleanup();
```
jeromecomte, line 330: Not sure if it is the best place to cleanup the staging resource manager but it does the job for the moment

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/AutomationPackageManager.java
```diff
@@ -627,6 +635,7 @@ protected static class Staging {
         List<Plan> plans = new ArrayList<>();
         List<ExecutiontTaskParameters> taskParameters = new ArrayList<>();
         List<Function> functions = new ArrayList<>();
+        ResourceManager resourceManager = new LocalResourceManagerImpl(new File("resources", new ObjectId().toString()));
```
jeromecomte, line 638: I would maybe use a dedicated context name "ap_staging_resources" for instance in order to quickly identify the root cause of potential leaks in the future

jeromecomte (approved):
Approving with 2 minor remarks

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/AutomationPackageManager.java
david-stephan: renamed as suggested in last commit

PR #190, opened 2024-01-16
SED-2731 Refactoring handling of ResourceManagers in context objects

jeromecomte:


step-core/src/main/java/step/resources/LayeredResourceManager.java
```diff
@@ -28,6 +28,8 @@
 
 public class LayeredResourceManager implements ResourceManager {
 
+
+    private final LayeredResourceAccessor layeredAccessor = new LayeredResourceAccessor();
```
david-stephan, line 32: @jeromecomte layeredAccessor doesn't seem to be used and should be removed.

david-stephan (changes requested):

david-stephan (approved):

PR #192, opened 2024-01-18
Sed 2734 plan annotation not supported for local execution of ap

david-stephan:


step-automation-packages/step-automation-packages-junit/src/test/java/step/client/StepRunnerYamlPlanTest.java
```diff
@@ -19,11 +19,16 @@
...
 import org.junit.runner.RunWith;
+import step.automation.packages.junit.AutomationPackagePlans;
 import step.handlers.javahandler.AbstractKeyword;
 import step.junit.runner.Step;
+import step.junit.runners.annotations.ExecutionParameters;
 import step.junit.runners.annotations.Plans;
 
 @RunWith(Step.class)
-@Plans({"composite-simple-plan.yml"})
+@Plans({"plans/composite-simple-plan.yml"})
```
iegorov777, line 29: @david-stephan 
1. Now we have to specify both `@AutomationPackagePlans` (to choose the plans to be executed in test) and `@Plans` (to configure plans to be scanned in AP Reader). It may be not clear for user. Maybe we can rename the `@AutomationPackagePlans` or provide some documentation to explain how it should be used?

2. Is it OK that `@Plans` is now not fully compatible to old versions (plans are now referenced basing on root folder but not on current package)?

step-automation-packages/step-automation-packages-support/src/main/java/step/automation/packages/AutomationPackageArchiveType.java
```diff
@@ -0,0 +1,8 @@
+package step.automation.packages;
+
+public enum AutomationPackageArchiveType {
```
iegorov777, line 3: @david-stephan 

I think there could be several different kinds of properties describing the AP type:
* contentType (i.e. language) - Java, DotNet, Undefined, etc.
* packagingType - jar (for java), classpath, zip, folder, ...
* the 'local' flag (to read keywords as LocalFunctions in `AutomationPackageKeyword#extractAnnotatedKeywords`)

So further we can redesign reading the AP in more or less general way:
* `AutomationPackageArchiveProvider` provides the `AutomationPackageArchive` basing on packagingType (unzips archive or prepares the URLClassLoader for jar or just references the source folder)
* `AutomationPackageArchive` contains the info about `contentType` and the `local` flag to be used by AP reader to scan the content (keywords and plans)

step-automation-packages/step-automation-packages-support/src/main/java/step/automation/packages/AutomationPackageArchive.java
```diff
@@ -39,25 +41,28 @@ public class AutomationPackageArchive implements Closeable {
     private final File originalFile;
     private final String originalFileName;
     private boolean internalClassLoader = false;
+    private final AutomationPackageArchiveType type;
 
     public AutomationPackageArchive(ClassLoader classLoader) {
         this.classLoader = classLoader;
         this.originalFile = null;
         this.originalFileName = null;
+        this.type = JAVA;
```
iegorov777, line 50: @david-stephan 
Shouldn't we also support the `type` field in yaml descriptor for AP?

step-automation-packages/step-automation-packages-junit/src/test/java/step/client/StepRunnerYamlPlanTest.java
```diff
@@ -19,11 +19,16 @@
...
+@AutomationPackagePlans(value = {"plans/composite-simple-plan.yml"})
 public class StepRunnerYamlPlanTest extends AbstractKeyword {
```
iegorov777, line 31: @david-stephan 
I noticed that the AP is read twice during local execution
* 1st time from `AutomationPackageKeywordsPlugin`
* 2nd time in `Step` runner

As I remember, the `AutomationPackageKeywordsPlugin` was initially designed to support use of keywords defined in current AP if we simply run a plan-text plan without AP junit runner. 

But after your last changes the old `Step` runner supports APs, so the `AutomationPackageKeywordsPlugin` probably can be deleted.

PR #199, opened 2024-01-25
SED-2594 extract controller dependcies from automation package manager module

david-stephan:


iegorov777 (approved):

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/AutomationPackageManager.java
cl-exense, line 1: License header is present twice in this file

david-stephan: cleaned up

PR #200, opened 2024-01-26
SED-2765 step-runner-local-execution-not-cleaning-up-resources-properly

david-stephan:


step-automation-packages/step-automation-package-controller/src/test/java/step/automation/packages/AutomationPackageManagerOSTest.java
```diff
@@ -108,6 +109,13 @@ public void onDelete(ExecutiontTaskParameters entity) {
         );
     }
 
+    @After
+    public void after() {
+        if (resourceManager != null) {
```
iegorov777, line 114: @david-stephan 
In default constructor for `LocalResourceManagerImpl` couldn't we mark the root directory to be automatically deleted on JVM exit via `FileUtils#forceDeleteOnExit`? This can solve potential issues if we forget to call explicit  `resourceManager.cleanup()` in some places.

Or it is to dangerous change, what do you think?

step-automation-packages/step-automation-packages-junit/src/main/java/step/engine/plugins/AutomationPackageKeywordsPlugin.java
```diff
@@ -50,6 +50,7 @@ public void initializeExecutionEngineContext(AbstractExecutionEngineContext pare
                                     new ObjectId(),
                                     context.require(FunctionTypeRegistry.class),
                                     functionAccessor,
+                                    resourceManager,
```
iegorov777, line 53: @david-stephan 
I forgot again, what the manager is used in execution engine context. I suppose, that in 100% this manager is prepared 
in `AbstractStepContext#setDefaultAttributes` (i.e `new LocalResourceManagerImpl(new File(contextPath() + "/resources")...`). 

So we have a guarantee, that the manager for isolated AP is always local?

iegorov777 (approved):

step-automation-packages/step-automation-package-controller/src/test/java/step/automation/packages/AutomationPackageManagerOSTest.java
david-stephan: @iegorov777 LocalResourceMaanger could potentially reuse an exsiting local folder, If think it is the responsability of the client to clean up the folder. However when creating a local resource manager like in the AP staging we could set the flag delete On exist, i.e. in 
ResourceManager resourceManager = new LocalResourceManagerImpl(new File("ap_staging_resources_" + new ObjectId()));
@jeromecomte what do you think?

step-automation-packages/step-automation-packages-junit/src/main/java/step/engine/plugins/AutomationPackageKeywordsPlugin.java
david-stephan: @iegorov777 the resource manager for execution context can be overriden, it is the case for executions in controller mode, here the plugin is only used in context of local execution, so this should be safe

PR #213, opened 2024-02-19
Sed 2810 updating ap change the resource i ds

iegorov777:


david-stephan:
Another approach was implemented to resolve this issue, closing this PR without merging

PR #222, opened 2024-02-28
SED-2899 Housekeeping of time series and measurement data is slow with PostgreSQL

david-stephan:


step-controller/step-controller-server/src/main/java/step/core/Controller.java
```diff
@@ -68,7 +68,7 @@
 
 public class Controller {
 
-	public static final Version VERSION = new Version(3,24,0);
+	public static final Version VERSION = new Version(3,24,3);
```
cl-exense, line 71: So this is how you smuggle in a version bump... :)

step-controller/step-controller-server/src/main/java/step/migration/tasks/FixPostgreSQLIndexes.java
```diff
@@ -0,0 +1,53 @@
...
+    public static final String FUNCTIONS_COLLECTION_PROPERTY = "functions";
+    public static final String PLANS_COLLECTION_PROPERTY = "plans";
+    public static final String TASKS_COLLECTION_PROPERTY = "tasks";
+
+    public FixPostgreSQLIndexes(CollectionFactory collectionFactory, MigrationContext migrationContext) {
+        super(new Version(3,24,3), collectionFactory, migrationContext);
+    }
+
+    @Override
+    public void runUpgradeScript() {
```
cl-exense, line 27: Do I understand correctly that the required new index are automatically created when needed? (and can not be accidentally re-deleted by this task)?

cl-exense (approved):
Approving, just a few non-critical questions...

step-controller/step-controller-server/src/main/java/step/core/Controller.java
david-stephan: indeed :-)

step-controller/step-controller-server/src/main/java/step/migration/tasks/FixPostgreSQLIndexes.java
david-stephan: The migration tasks is only executed once (first time started with 3.24.3, the creation of indexes is happening later in the startup lifecyle. We always try to create the index at startup if they do not exist already

PR #229, opened 2024-03-07
SED-2942 fixing screen templates

jeromebrongniart:


step-controller/step-controller-server/src/main/java/step/plugins/screentemplating/ScreenTemplateService.java
```diff
@@ -98,7 +103,7 @@ public List<Input> getInputsForScreenGet(@PathParam("id") String screenId, @Cont
 	public List<Input> getInputsForScreenPost(@PathParam("id") String screenId, Object params) {
 		ObjectPredicate objectPredicate = objectPredicateFactory.getObjectPredicate(getSession());
 		Map<String, Object> contextBindings = getContextBindings(null);
-		if(params != null && params instanceof Map) {
+		if(params instanceof Map) {
```
david-stephan, line 106: @jeromebrongniart We should keep the check on null

david-stephan (changes requested):

david-stephan (approved):

PR #231, opened 2024-03-07
SED-2944 More robust OryonFunction Unit Tests (prerequisite)

cl-exense:


jeromecomte (changes requested):
Declining this PR. No validation should be done in the model. This model class could be instantiated on the client side or somewhere where the file doesn't exist and this is perfectly fine

jeromecomte:
@cl-exense Short amendment to my previous comment: The validation itself makes sense. If required this could be done in the FunctionType class overriding following methods: step.functions.type.AbstractFunctionType#updateFunction / setupFunction

PR #233, opened 2024-03-24
SED-2708-added refreshInterval to dashboard model

lucianbaciu:


david-stephan (approved):
@lucianbaciu approving assuming you support null value in the FE (the value will be null for dashboard saved with previous version).

PR #234, opened 2024-03-27
SED-2852-switch-sleep-control

david-stephan:


cl-exense:
Approving, the only thing that should be corrected is the comment describing what the migration task does because it does not correspond to the implementation. Otherwise looks excellent.

cl-exense (dismissed):

PR #236, opened 2024-04-04
Sed 2979 fixing exclude annotations

jeromebrongniart:


step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/AbstractAutomationPackageReader.java
```diff
@@ -282,19 +282,22 @@ public static List<Plan> extractAnnotatedPlans(AutomationPackageArchive archive,
             }
         }
 
+        Set<Method> excludedMethods = new HashSet<>();
+
         // Classes with @Plan annotation in methods
         // and filter them
         for (Method m : annotationScanner.getMethodsWithAnnotation(step.junit.runners.annotations.Plan.class)) {
-            if (!excludedByAnnotation.contains(m.getDeclaringClass()) && !classesWithPlans.contains(m.getDeclaringClass())) {
+//            if (!excludedByAnnotation.contains(m.getDeclaringClass()) && !classesWithPlans.contains(m.getDeclaringClass())) {
```
david-stephan, line 290: @jeromebrongniart if this codes is now working, we should remove the comments before merging. and fix the indentation. I think however that we could keep the if (!excludedByAnnotation.contains(m.getDeclaringClass()), since all methods of such class will be ignored anyway.

david-stephan (changes requested):

david-stephan (approved):

PR #238, opened 2024-04-09
Sed 2874 plans refactor ap for easier intregration of new entities

iegorov777:


step-core/src/main/java/step/core/yaml/serializers/StepYamlSerializersScanner.java
```diff
@@ -0,0 +1,77 @@
...
+public class StepYamlSerializersScanner {
+
+    private static final Logger log = LoggerFactory.getLogger(StepYamlSerializersScanner.class);
+
+    /**
+     * Scans and returns all {@link StepYamlSerializer} classes annotated with {@link StepYamlSerializerAddOn}
+     */
+    public static List<SerializerBind<?>> scanSerializerAddons(ObjectMapper yamlObjectMapper) {
+        List<SerializerBind<?>> result = new ArrayList<>();
+        List<Class<?>> annotatedClasses = new ArrayList<>(CachedAnnotationScanner.getClassesWithAnnotation(StepYamlSerializerAddOn.LOCATION, StepYamlSerializerAddOn.class, Thread.currentThread().getContextClassLoader()));
```
jeromecomte, line 42: Same comment as above. We might replace this by explicit registration. To be discussed

step-plans/pom.xml
```diff
@@ -35,7 +35,8 @@
 		<module>step-plans-yaml-parser</module>
 		<module>step-plans-runners</module>
 		<module>step-plans-base-artefacts</module>
-	</modules>
+        <module>step-plans-yaml-support</module>
```
jeromecomte, line 38: Formatting

step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/AfterSequence.java
```diff
@@ -18,8 +18,12 @@
  ******************************************************************************/
 package step.artefacts;
 
+import step.automation.packages.AutomationPackageNamedEntity;
 import step.core.artefacts.Artefact;
+import step.plans.parser.yaml.model.YamlModel;
 
+@YamlModel
+@AutomationPackageNamedEntity(name = "afterSequence")
```
jeromecomte, line 26: Can we have a AutomationPackageNamedEntity which is not a YamlModel? i.e. do we need both annotations?

step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/automation/YamlFunctionGroup.java
```diff
@@ -0,0 +1,63 @@
...
+    private DynamicValue<String> containerCommand = new DynamicValue<>();
+
+    public YamlFunctionGroup() {
+        super(FunctionGroup.class);
+    }
+
+    @Override
+    protected void fillArtefactFields(FunctionGroup res) {
+        super.fillArtefactFields(res);
+        if (dockerImage != null) {
```
jeromecomte, line 39: Can't we do these not null checks in a generic manner?

jeromecomte (commented):
- General code quality looks good. Code looks much easier to understand than before. The grouping of the schema generation and serialization logic is also an improvement.
- Added comment regarding registry of extensions that should be discussed. 
- Mapping of Yaml model to model should be discussed (see comment about get null checks)
- Test coverage to be checked.
- Validation of the implementation for DB model changes to be done (details follow)

step-plans/pom.xml
iegorov777: @jeromecomte 
Fixed

step-automation-packages/step-automation-packages-schema/src/main/java/step/automation/packages/yaml/schema/YamlKeywordSchemaGenerator.java
```diff
@@ -50,30 +49,29 @@ public YamlKeywordSchemaGenerator(JsonProvider jsonProvider) {
...
 
         // --- Fields metadata rules (fields we want to rename)
-        FieldMetadataExtractor fieldMetadataExtractor = prepareMetadataExtractor();
+        FieldMetadataExtractor fieldMetadataExtractor = new DefaultFieldMetadataExtractor();
 
         List<JsonSchemaFieldProcessor> processingRules = prepareFieldProcessors();
         this.jsonSchemaCreator = new JsonSchemaCreator(jsonProvider, new AggregatedJsonSchemaFieldProcessor(processingRules), fieldMetadataExtractor);
     }
 
+    protected List<JsonSchemaDefinitionCreator> getDefinitionsExtensions() {
```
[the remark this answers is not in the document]
iegorov777: @jeromecomte 
The idea looks good, but we need to discuss how to do it in the best way. As I see, there are 2 separate issues:

1. The json schema generation. 
In fact, these scanned `JsonSchemaDefinitionCreators` are only used in `YamlPlanJsonSchemaGenerator` and `YamlAutomationPackageSchemaGenerator`. The generation is called from standalone tools (`YamlPlanSchemaGenerationTool`) and doesn't use plugins engine.  So I'm not sure that plugins is the best place to register such extensions for json schema.  Alternatively to avoid scanning we could simply register all creators manually inside the `YamlPlanJsonSchemaGenerator`, but in this case we should not forget to do this every time when we need to extend the schema. So I still don't know, which approach is better...

2. The extensions for runtime (like custom serializers/deserializers), which really could be registered in plugins. But there are some very common serializers (like `YamlDynamicValueSerializer`), which should be always used (explicitly listed in `YamlPlanReader`?). Also we need to register these extensions not only in controller plugins, but also for junit runners and in tests.

FYI. I've checked the slowest annotation scans in Step and found the following:
* the `ExecutionEngine.withPluginsFromClasspath()` used in junit runner
* the `AnnotationScanner.forAllClassesFromClassLoader(archive.getClassLoader())` in `AutomationPackageReader` used to scan keywords and plans
Other ones (like for `StepYamlDeserializerAddOn`) are much faster, because they are restricted with 'step' package and don't scan the whole classpath.

step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/automation/YamlFunctionGroup.java
iegorov777: @jeromecomte 
I'll try to write some method to check/copy all fields in model by reflection. Then all this checks and copying will be perfomed implicitly.

step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/AfterSequence.java
iegorov777: @iegorov777 
Agree. I think we can only the `YamlModel` annotation (as more general) and add the the new `yamlName` propery to   `YamlModel` to define the name.

step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/automation/YamlFunctionGroup.java
iegorov777: @jeromecomte @david-stephan 

I've added the `AbstractYamlModel` class with `copyFieldsToObject` and `copyFieldsFromObject` methods to copy all fields automatically. If the custom logic is required for some field, this field should be marked with new `@YamlFieldCustomCopy`. 
The new solution is applied in all artefact models now.

step-automation-packages/step-automation-packages-schema/src/main/java/step/automation/packages/yaml/schema/YamlKeywordSchemaGenerator.java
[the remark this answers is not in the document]
david-stephan: @iegorov777 Thanks for the analysis, I'll created a dedicated ticket for next release to see if there are any improvement to be done. For this iterations, we can keep the implementation as it is.

step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/AfterSequence.java
david-stephan: @iegorov777 agreed

step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/automation/YamlFunctionGroup.java
david-stephan: @iegorov777 looks good

step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/AfterSequence.java
iegorov777: @david-stephan 
Implemented

david-stephan (approved):

PR #240, opened 2024-04-10
SED-2921-add-name-labels-to-prometheus-metrics

david-stephan:


step-controller-plugins/step-controller-measurement/src/main/java/step/plugins/measurements/Measurement.java
```diff
@@ -85,4 +112,14 @@ public void addCustomFields(Map<String, ?> fields) {
 		this.putAll(fields);
 	}
 
+	public Map<String, String> getAdditionalAttributes() {
+		return additionalAttributes;
+	}
+
+	public void addAdditionalAttributes(TreeMap<String, String> additionalAttributes) {
```
cl-exense, line 119: Is there a reason that this must be a TreeMap? In general, methods should accept the generic collection type unless there is a specific need for a concrete implementation.

david-stephan: For the logic it's important that the map is ordered (also the one provided as input) ), I just want to enforce it with that.

PR #242, opened 2024-04-11
Sed 2966 table dashlet for custom dashboards

lucianbaciu:


lucianbaciu:
@david-stephan currently I use a basic normalized (flat) structure for chart/table settings. If you want we can also play with custom serialization, but ideally we would should keep the FE structure simple

david-stephan (changes requested):
@lucianbaciu the PR looks OK, however this we are chaning the model I assume dashboard created in version 24 won't be compatible anymore with Step 25, therefore we must implement a migration Task as part of this change. See MigrateArtefactsToPlans for an example, and use MigrationManager to register it.

lucianbaciu:
@david-stephan,
`MigrationManager migrationManager = context.get(MigrationManager.class);`
I do see the class, but it returns null. do I miss any depdendency?

step-controller-plugins/step-controller-plugins-timeseries/pom.xml
```diff
@@ -55,6 +55,12 @@
             <version>${step-framework.version}</version>
             <scope>test</scope>
         </dependency>
+         <dependency>
```
david-stephan, line 58: @lucianbaciu  why do we need this dependency?

david-stephan:
@lucianbaciu you must ensure that the migration manager is available in the context at the time you try to get it. First please replace context.get by context.require and add the required plugin dependency (@Plugin(dependencies= {MigrationManagerPlugin.class}))

step-controller-plugins/step-controller-plugins-timeseries/pom.xml
lucianbaciu: @david-stephan all tests were failing with an exception coming from that package. after adding it, it worked. now I removed it, but everything still works, so I will clean it. we'll see in the build if something is not ok

david-stephan (dismissed):

david-stephan (approved):

PR #243, opened 2024-04-11
SED-3008-Ability to specify colorization type for charts (ONLY_STROKE…

lucianbaciu:
… or FILL)


step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/dashboards/model/AxesSettings.java
```diff
@@ -18,6 +18,8 @@ public class AxesSettings {
 	@NotNull
 	private String unit;
 	private MetricRenderingSettings renderingSettings;
+	
+	private AxesColorizationType colorizationType;
```
david-stephan, line 22: @lucianbaciu same remark as for the other PR, can you clarify if a migration task is required? When migrating from Ste 24 colorizationType will be null in DB, is that supported? Otherwise instead of a mgiration tasks for this one, a default value could be defined with private AxesColorizationType colorizationType = AxesColorizationType.STROKE;

lucianbaciu: @david-stephan for this task no, migration is not required. I will do the change proposed by you (I like it better instead of handling it in FE)

david-stephan (approved):

PR #244, opened 2024-04-12
SED-3018 modification-of-global-entities-from-another-project-should-…

david-stephan:
…not-be-allowed

### DIFF
--- a/step-controller/step-controller-server/src/main/java/step/core/deployment/AbstractStepServices.java
+++ b/step-controller/step-controller-server/src/main/java/step/core/deployment/AbstractStepServices.java
@@ -22,13 +22,12 @@ import jakarta.annotation.PostConstruct;
 import jakarta.inject.Inject;
 
 import ch.exense.commons.app.Configuration;
+import org.apache.http.HttpStatus;
 import step.core.GlobalContext;
 import step.core.access.User;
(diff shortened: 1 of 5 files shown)


step-controller/step-controller-server/src/main/java/step/core/deployment/AbstractStepServices.java
```diff
@@ -96,4 +95,19 @@ protected void checkRightsOnBehalfOf(String right, String userOnBehalfOf) {
 			throw new AuthorizationException(ex.getMessage());
 		}
 	}
+
+	/**
+	 * The ObjectHookInterceptor.aroundReadFrom can only be used for POST request passing an entity as request BODY
+	 * This method can be used as helper for all other cases where checking if the entity is acceptable in given context (i.e. DELETE request...(
+	 * @param entity the entity to be asserted
+	 */
+	protected void assertEntityIsAcceptableInContext(AbstractIdentifiableObject entity) {
```
jeromecomte, line 104: I would maybe change the name of this helper method to something more meaningful like: isObjectEditableInContext

jeromecomte (approved):
PR looks fine. Added minor remark. E2E REST tests should be mandatory for these security relevant features

PR #248, opened 2024-04-16
SED-3024 implemented new flags to disable counts and object enrichment in table service

jeromecomte:


step-controller/step-controller-remote-client/src/main/java/step/client/collections/remote/RemoteCollection.java
```diff
@@ -176,7 +176,13 @@ protected UnsupportedOperationException notImplemented()  {
         return new UnsupportedOperationException("This method is currently not implemented");
     }
 
-	@Override
+    @Override
+    // Empty implementation to be replaced
```
cl-exense, line 180: I don't understand this change or the comment :-p

cl-exense (commented):
Not sure what the getName() change is good for, please clarify.

step-controller/step-controller-remote-client/src/main/java/step/client/collections/remote/RemoteCollection.java
jeromecomte: Hehe, I had indeed problem to explain it in a few words

PR #250, opened 2024-04-26
SED-3038-limit-number-of-series

david-stephan:


lucianbaciu (approved):
It looks good and simple. We may improve also the lower side which fetches the data, but if you didn't notice perf issues, it doesn't make sense for now.

david-stephan:
Thanks for the feedback. Since all data are fetched from the DB and grouped in memory, limiting the number of groups (i.e. series) here would not improve the perf much. We could consider limiting the overall number of buckets (across series) to prevent OOM, but this didn't seem to be a risk as of now.

PR #251, opened 2024-04-29
Sed 2998 support step paramters in ap

iegorov777:


step-automation-packages/step-automation-packages-controller/src/main/java/step/parameter/automation/AutomationPackageParameterPlugin.java
```diff
@@ -0,0 +1,49 @@
...
+import step.core.plugins.Plugin;
+import step.parameter.Parameter;
+import step.plugins.parametermanager.ParameterManagerControllerPlugin;
+
+@Plugin(dependencies= {AutomationPackagePlugin.class, ParameterManagerControllerPlugin.class})
+public class AutomationPackageParameterPlugin extends AbstractControllerPlugin {
+
+    @Override
+    public void serverStart(GlobalContext context) throws Exception {
+        registerParametersHooks(context.require(AutomationPackageHookRegistry.class), context.require(AutomationPackageSerializationRegistry.class), (Accessor<Parameter>) context.get("ParameterAccessor"));
```
david-stephan, line 39: I assume the accessor is required too

step-automation-packages/step-automation-packages-support/src/main/java/step/automation/packages/model/AutomationPackageSchedule.java
```diff
@@ -16,13 +16,15 @@
  * You should have received a copy of the GNU Affero General Public License
  * along with STEP.  If not, see <http://www.gnu.org/licenses/>.
  ******************************************************************************/
-package step.plans.parser.yaml.automation.model;
+package step.automation.packages.model;
```
david-stephan, line 19: I would have move it next to the DB model, i.e. step-plan-core

step-automation-packages/step-automation-packages-support/src/main/java/step/automation/packages/model/AutomationPackageParameter.java
```diff
@@ -0,0 +1,89 @@
...
+
+import step.commons.activation.Expression;
+import step.core.yaml.AbstractYamlModel;
+import step.core.yaml.YamlFieldCustomCopy;
+import step.core.yaml.YamlModel;
+import step.parameter.Parameter;
+import step.parameter.ParameterScope;
+
+@YamlModel(named = false)
+public class AutomationPackageParameter extends AbstractYamlModel {
```
david-stephan, line 29: I would have created it next to the DB model, i.e. step-core-model

step-automation-packages/step-automation-packages-support/src/main/java/step/automation/packages/deserialization/AutomationPackageParametersRegistration.java
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with STEP.  If not, see <http://www.gnu.org/licenses/>.
+ ******************************************************************************/
+package step.automation.packages.deserialization;
+
+import step.automation.packages.model.AutomationPackageParameter;
+import step.automation.packages.model.AutomationPackageSchedule;
+
+public class AutomationPackageParametersRegistration {
```
david-stephan, line 24: I see this class is also used in Junit, I assume that's why it is not directly in the controller plugin? Also how does it work for the Step runner? I'm also not sure whether this is the right location.

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/hooks/ExecutionTaskParameterWithoutSchedulerHook.java
```diff
@@ -16,23 +16,23 @@
  * You should have received a copy of the GNU Affero General Public License
  * along with STEP.  If not, see <http://www.gnu.org/licenses/>.
  ******************************************************************************/
-package step.core.scheduler;
+package step.automation.packages.hooks;
```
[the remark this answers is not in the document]
iegorov777: @david-stephan 

The main scheduler (using scheduler) hook is still in scheduler plugin. But I moved the `ExecutionTaskParameterWithoutSchedulerHook` from the controller module to support at least reading the 'schedules` section defined in automation package yaml in tests. Othewise the `schedules` section will be completely ignored in all modules not relying on `controller` module.

As we discussed earlier, it is OK to keep the hook only in controller plugins (and make all tests dependent on contoller module). But while implementing the same for `Parameters` I've doubted whether it is always fine solution. In this case for parameters we also will ignore 'parameters' section in automation package reader and manager. So I thought that maybe it is more flexible to register the hook still in contoller plugin, but keep the hook class outside of controller module to reuse it in local runs or in tests if it is required.

So the main question again is how to split the entities and logic between maven modules in the most convenient way.

step-automation-packages/step-automation-packages-controller/src/main/java/step/parameter/automation/AutomationPackageParameterPlugin.java
iegorov777: @david-stephan 
Indeed. Fixed

david-stephan (approved):

david-stephan (approved):

david-stephan (approved):

david-stephan (approved):

PR #261, opened 2024-05-11
SED-1012: support stateless REST calls (tenants)

iegorov777:


step-maven-plugin/src/main/java/step/plugins/maven/ExecuteAutomationPackageMojo.java
```diff
@@ -200,16 +190,20 @@ protected ControllerCredentials getControllerCredentials() {
...
     }
 
-    protected MultitenancyClient createMultitenancyClient() {
-        return new RemoteMultitenancyClientImpl(getControllerCredentials());
-    }
-
     protected RemoteAutomationPackageClientImpl createRemoteAutomationPackageClient() {
-        return new RemoteAutomationPackageClientImpl(getControllerCredentials());
+        RemoteAutomationPackageClientImpl client = new RemoteAutomationPackageClientImpl(getControllerCredentials());
+        addProjectHeaderToRemoteClient(client);
```
jeromecomte, line 195: To avoid the need to call the method addProjectHeaderToRemoteClient after each instantiation, I was wondering if it would make sense to add the field project or a generic map for the context attributes to the controller credentials

jeromecomte (approved):

step-controller/step-controller-multitenancy-model/src/main/java/step/controller/multitenancy/Constants.java
```diff
@@ -0,0 +1,23 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with STEP.  If not, see <http://www.gnu.org/licenses/>.
+ ******************************************************************************/
+package step.controller.multitenancy;
+
+public interface Constants {
+    String TENANT_HEADER = "StepProject";
```
david-stephan, line 22: @iegorov777 It should not be camel case but Step-Project

step-maven-plugin/src/main/java/step/plugins/maven/ExecuteAutomationPackageMojo.java
iegorov777: @jeromecomte 
Not sure that the ControllerCredentials is the best place to keep the map with additional headers. But I've changed the simple map in the `AbstractRemoteClient` to the wrapping `AdditionalHeaders` to add some default headers (tenant header) easier.

step-controller/step-controller-multitenancy-model/src/main/java/step/controller/multitenancy/Constants.java
iegorov777: @david-stephan 
Renamed

PR #264, opened 2024-05-13
SED-2992-allow-to-duplicate-analytics-dashboards

david-stephan:


step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/migration/MigrateDashboardsTask.java
```diff
@@ -28,23 +28,45 @@ public MigrateDashboardsTask(CollectionFactory collectionFactory, MigrationConte
...
+
+    private void migrateNamePropertyToAttributes() {
+        dashboardsCollection.find(Filters.empty(), null, null, null, 0).forEach(dashboard -> {
+            DocumentObject attributes = dashboard.getObject("attributes");
+            if (attributes == null) {
+                attributes = new DocumentObject();
+                dashboard.put("attributes", attributes);
+            }
+            String name = (String) dashboard.remove("name");
+            attributes.put("name", name);
```
lucianbaciu, line 48: I'd use AbstractOrganizableObject.NAME if we have it

lucianbaciu (approved):

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/migration/MigrateDashboardsTask.java
lucianbaciu: nevermind. we should hardcode values in migration in case the variables will change later

PR #266, opened 2024-05-14
SED-3086 automation package and assertion plan execution stability issues25

david-stephan:


step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/AutomationPackageExecutionPlugin.java
```diff
@@ -30,40 +32,68 @@ public AutomationPackageExecutionPlugin(AutomationPackageLocks automationPackage
...
     }
 
     @Override
-    public void afterExecutionEnd(ExecutionContext context) {
-        super.afterExecutionEnd(context);
-        String automationPackageId = getAutomationPackageId(context);
-        if (automationPackageId != null) {
+    public void executionFinally(ExecutionContext context) {
+        super.executionFinally(context);
+        if (shouldLock(context)) {
```
jeromecomte, line 89: If for some reason the result of shouldLock would be different from the call in "executionStart" we would never release the lock. It might be safer to store the information that the lock has to be released within the context to be sure that we release it here

step-controller/step-controller-server/src/main/java/step/core/scheduler/Executor.java
```diff
@@ -64,9 +64,15 @@ public Executor(GlobalContext globalContext) {
 		
 		try {
 			Properties props = getProperties();
-			schedulerFactory = new StdSchedulerFactory(props);
+			StdSchedulerFactory schedulerFactory = new StdSchedulerFactory(props);
 			scheduler = schedulerFactory.getScheduler();
 			scheduler.setJobFactory(new ExecutionJobFactory(globalContext, executionEngine));
+			//Create another scheduler with same number of threads for nested executions to avoid deadlock
+			Properties propsNested = getProperties();
```
jeromecomte, line 71: These properties shouldn't be different from the props above

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/AutomationPackageExecutionPlugin.java
david-stephan: I changed the implementation based on your comment

david-stephan:
@jeromecomte ready for another review

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/AutomationPackageExecutionPlugin.java
```diff
@@ -30,40 +33,70 @@ public AutomationPackageExecutionPlugin(AutomationPackageLocks automationPackage
...
                 boolean locked = automationPackageLocks.tryReadLock(automationPackageId);
                 if (!locked) {
-                    logger.debug("Timeout while acquiring read lock on automation package " + automationPackageId);
+                    debugLog(context, "Timeout while acquiring read lock on automation package.");
                     throw new PluginCriticalException("Timeout while acquiring lock on automation package with id " +
                             automationPackageId + ". This usually means that an update of this automation package is on-going and took more than the property " +
                             AUTOMATION_PACKAGE_READ_LOCK_TIMEOUT_SECS + " (default " + AUTOMATION_PACKAGE_READ_LOCK_TIMEOUT_SECS_DEFAULT + " seconds)");
                 }
-                logger.debug("Acquired read lock on automation package " + automationPackageId);
+                context.put(EXECUTION_CONTEXT_LOCK_ID, automationPackageId);
```
jeromecomte, line 47: it doesn't change anything in the code flow here as we throw an exception in the if-clause but I would generally recommend to put this in the else block

jeromecomte (approved):

PR #276, opened 2024-06-06
SED-3105-added missing fields

lucianbaciu:


david-stephan (changes requested):
We're missing the changes regarding the prepopulated dashboards (initialization/migration)

david-stephan (approved):

PR #277, opened 2024-06-08
Sed 2870 cli for automation packages

iegorov777:


step-maven-plugin/src/test/java/step/plugins/maven/ExecuteAutomationPackageMojoTest.java
```diff
@@ -19,159 +19,45 @@
...
 		mojo.execute();
-
-		assertAutomationPackageClientMockCalls(remoteAutomationPackageClientMock);
-	}
-
-	private static List<String> getExecuteAutomationPackageResult(List<Execution> executions) {
-		return executions.stream().map(e -> e.getId().toString()).collect(Collectors.toList());
-	}
-
-	@Test
```
david-stephan, line 72: @iegorov777 I've the feeling we now only have one happy path tests and removed all other negatives test (import error, execution failed..., is that expected

david-stephan (approved):

PR #279, opened 2024-06-19
SED-3164 customizable percentiles in charts and tables

lucianbaciu:


step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/dashboards/DashboardsGenerator.java
```diff
@@ -264,7 +264,19 @@ private static DashboardItem createThreadGroupDashlet(MetricType metric) {
...
+                new TableDashletSettings.ColumnSelection(TableChartColumn.AVG, true),
+                new TableDashletSettings.ColumnSelection(TableChartColumn.MIN, true),
+                new TableDashletSettings.ColumnSelection(TableChartColumn.MAX, true),
+                new TableDashletSettings.ColumnSelection(TableChartColumn.PCL_80, 80D, true),
+                new TableDashletSettings.ColumnSelection(TableChartColumn.PCL_90, 90D, true),
+                new TableDashletSettings.ColumnSelection(TableChartColumn.PCL_99, 99D, true),
+                new TableDashletSettings.ColumnSelection(TableChartColumn.TPS, true),
+                new TableDashletSettings.ColumnSelection(TableChartColumn.TPH, true)
+        );
+//        return Arrays.stream(TableChartColumn.values()).map(c -> new TableDashletSettings.ColumnSelection(c, true)).collect(Collectors.toList());
```
david-stephan, line 279: commented code should be cleaned up

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/dashboards/model/TableDashletSettings.java
```diff
@@ -22,6 +21,8 @@ public TableDashletSettings setColumns(List<ColumnSelection> columns) {
     public static class ColumnSelection {
         @NotNull
         private TableChartColumn column;
+        
+        private Double pclValue; // can override when column is PCL
```
david-stephan, line 25: You should extend this class for PclColumnSelection

david-stephan (changes requested):
see individual comments

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/dashboards/model/TableDashletSettings.java
lucianbaciu: @david-stephan I just tried to see how the FE lib generator would work with that, and I think this would create too much overhead. It seems the subclass is not recognized, which means I should cast and force properties access all over the place. Let me know if I miss something.

lucianbaciu: @david-stephan I talked with Tim this morning and there is no easy or automated way to handle these things.

PclColumnSelection is not generated because it is nowhere declared in the API (it's just a subclass of the declared one).
We can manually create this class in the FE, extend the generated one. The bad part is that we still have to cast our ColumnSelection all over the place in order to access it's pclValue attribute. We could have this issue in the backend too.
To conclude, @neogucky  is fine with both approaches (keep the way it is with pclValue being optional, or apply your suggestion and create the subclass manually in FE), while I'd go for the first one. You can decide

david-stephan: @lucianbaciu we have class hierarchy all over the Step model and used as part of the REST service request and response payloads. There is no need of casting. You must use appropriate annotations in the BE to enable serialization/deserialization to the proper classes and generate the openapi JSON schema including the sub classes. 
Example:
`    @JsonSubTypes({
            @JsonSubTypes.Type(value = PercentileColumnSelection.class)
    })
    @JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, property = "class")
    public static class ColumnSelection {`
If still not clear, let's discuss it in our weekly

david-stephan (dismissed):

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/dashboards/model/TableDashletSettings.java
lucianbaciu: @david-stephan a migration script is needed because the saved columns don't have the "type" property saved in mongo, so the deserialization crashes. do you know any way so I can avoid it?

lucianbaciu:
@david-stephan I'm waiting to replace 0.0.0-SNAPSHOT framework but build server is not working.

david-stephan (approved):

neogucky (approved):

PR #287, opened 2024-07-05
SED-3265 execution-tree-node-shows-wrong-names

david-stephan:


jeromecomte (dismissed):
Approving as a workaround. In my opinion we shouldn't change the value of the attribute map which is meant to be static. We could check if the FE shouldn't use ReportNode.name or AbstractArtefact.dynamicName to retrieve the resolved value of the name

PR #291, opened 2024-07-15
SED-3273 Import and export of hidden plans incorrect behaviour

david-stephan:


step-controller/step-controller-server/src/main/java/step/core/plans/PlanPlugin.java
```diff
@@ -123,5 +128,18 @@ private void createTableSettingsIfNecessary(GlobalContext context, ScreenInput n
 		}
 	}
 
+	public static class PlanExportBiConsumer implements BiConsumer<Object, ExportContext> {
```
cl-exense, line 131: Rename this class to a more meaningful name, maybe like MakeInvisiblePlansVisibleExportHook or similar

step-core/src/main/java/step/core/entities/EntityManager.java
```diff
@@ -91,9 +94,10 @@ public void getEntitiesReferences(String entityType, ObjectPredicate objectPredi
 		if (entity == null ) {
 			throw new RuntimeException("Entity of type " + entityType + " is not supported");
 		}
-		entity.getAccessor().getAll().forEachRemaining(a -> {
+		Filter exportAllFilters = Objects.requireNonNullElse(getExportAllFilters(entityType), Filters.empty());
+		entity.getAccessor().getCollectionDriver().find(exportAllFilters,null,null,null,0).forEach(a -> {
```
cl-exense, line 98: This could really benefit from a short comment explaining what it does and why (i.e. what do the conditions mean)

cl-exense (changes requested):
Code as such is ok I think, but a few changes could make it easier to understand.

step-core/src/main/java/step/core/entities/EntityManager.java
david-stephan: added comments

step-controller/step-controller-server/src/main/java/step/core/plans/PlanPlugin.java
david-stephan: Renamed the class but still keeping it bit generic since this is not reserved to plan visibility and could be extended with further modifications of plan during export

PR #294, opened 2024-07-23
SED-3026-parameters-support-values-with-groovy-expression

david-stephan:


step-core/src/main/java/step/parameter/ParameterManager.java
```diff
@@ -34,37 +31,79 @@
...
+
+		ParameterScope scope = newParameter.getScope();
+		if(scope != null && scope.equals(ParameterScope.GLOBAL) && newParameter.getScopeEntity() != null) {
+			throw new ParameterManagerException("Scope entity cannot be set for parameters with GLOBAL scope.");
+		}
+
+		if(sourceParameter != null) {
+			// the parameter has been updated but the value hasn't been changed
+			DynamicValue<String> newParameterValue = newParameter.getValue();
+			if(newParameterValue != null && !newParameterValue.isDynamic() && newParameterValue.get().equals(PROTECTED_VALUE)) {
```
jeromecomte, line 82: Not sure if this is still the optimal implementation. It forces the clients to set the value to "******" in order for the value not to change when updating a parameter.

jeromecomte (commented):
PR looks good. Having only one comment. Would approve otherwise

step-core/src/main/java/step/parameter/ParameterManager.java
david-stephan: @jeromecomte so you would simply remove this if block and move the line of code 
```
newParameter.setValue(sourceParameter.getValue());
```
to the following if bloc 
```
if(isProtected(sourceParameter)) {
    // protected value should not be changed
    newParameter.setProtectedValue(true);
    newParameter.setValue(sourceParameter.getValue());
}
```

jeromecomte (approved):

PR #298, opened 2024-07-31
SED-2949 new Notifications

cl-exense:


step-core/src/main/java/step/automation/packages/AutomationPackageOperationMode.java
```diff
@@ -0,0 +1,7 @@
+package step.automation.packages;
+
+public enum AutomationPackageOperationMode {
+    MAIN,
```
jeromecomte, line 4: I would add a javadoc to document these 3 modes. Not Sure about the meaning of "MAIN"...

jeromecomte (changes requested):
Requesting minor change

step-core/src/main/java/step/automation/packages/AutomationPackageOperationMode.java
cl-exense: Hehe, yes! "MAIN" was just a fallback name because I didn't find a better one. The actual meaning is something like "PRODUCTION", or "REALLY_DO_IT" (as opposed to the other two), but these seemed even less appropriate, so I tried to use something generic. Any suggestion for the correct name? :)

PR #299, opened 2024-08-02
SED-2834-redesign-the-execution-report-views

neogucky:


step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/ReportNodeAccessorImpl.java
```diff
@@ -92,6 +121,15 @@ public Stream<ReportNode> getReportNodesByExecutionIDAndClass(String executionID
 						Filters.equals("_class", class_))),
 				new SearchOrder("executionTime", 1), null, null, 0);
 	}
+
+	@Override
+	public Stream<ReportNode> getReportNodesByExecutionIDAndClass(String executionID, String class_, Integer limit) {
+		assert executionID != null;
```
cl-exense, line 127: same as above

step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/aggregated/AggregatedReportView.java
```diff
@@ -0,0 +1,39 @@
+package step.core.artefacts.reports.aggregated;
+
+import step.core.artefacts.AbstractArtefact;
+
+import java.util.List;
+import java.util.Map;
+
+public class AggregatedReportView {
+
+    public final AbstractArtefact artefact;
```
cl-exense, line 10: Yay, immutable classes without getters and setters! (I'm serious, I like this). No need to change anything here.

step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/aggregated/AggregatedReportView.java
```diff
@@ -0,0 +1,39 @@
...
+    @Override
+    public String toString() {
+        StringBuilder stringBuilder = new StringBuilder();
+        recursiveToString(this, 0, stringBuilder);
+        return stringBuilder.toString();
+    }
+
+    private void recursiveToString(AggregatedReportView node, int level, StringBuilder stringBuilder) {
+        String name = node.artefact.getClass().getSimpleName();
+        String indentation = new String(new char[level]).replace("\0", " ");
```
cl-exense, line 35: That's pretty smart :-) But there's the even better (Java 11+) `String indentation = " ".repeat(level);`.

cl-exense: Probable bug: You are using name, countTotal() etc. from `this`, not `node`. Make recursiveToString static to avoid this class of bug.

step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/aggregated/AggregatedReportViewBuilder.java
```diff
@@ -0,0 +1,51 @@
...
+    public AggregatedReportView buildAggregatedReportView(AggregatedReportViewRequest request) {
+        Objects.requireNonNull(request);
+        Execution execution = executionAccessor.get(executionId);
+        String aggregatedReportRoot = execution.getResolvedPlanRootNodeId();
+        ResolvedPlanNode resolvedPlanNode = resolvedPlanNodeAccessor.get(aggregatedReportRoot);
+        return recursivelyBuildAggregatedReportTree(resolvedPlanNode, request);
+    }
+
+    public static class AggregatedReportViewRequest {
+        ReportNodeTimeSeries.Range range;
```
cl-exense, line 41: This should probably be public, not `<default>` . I would also make it final for good measure. (Yes, that will require an explicit constructor).

step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/aggregated/ReportNodeTimeSeries.java
```diff
@@ -0,0 +1,64 @@
...
+        reportNodeTimeSeries.createIndexes(Set.of(new IndexField(EXECUTION_ID, Order.ASC, String.class)));
+    }
+
+    public void ingestReportNode(ReportNode reportNode) {
+        ReportNodeStatus status = reportNode.getStatus();
+        timeSeriesIngestionPipeline.ingestPoint(new BucketAttributes(Map.of(EXECUTION_ID, reportNode.getExecutionID(), ARTEFACT_HASH, reportNode.getArtefactHash(), STATUS, status.toString())), System.currentTimeMillis(), 1);
+    }
+
+    public static class Range {
+        long from;
```
cl-exense, line 45: Again, public static classes should usually also have public fields. Also suggest using final unless they're explicitly meant to be mutable, and make nullability assumptions explicit.

step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/aggregated/ReportNodeTimeSeries.java
```diff
@@ -0,0 +1,64 @@
...
+import step.core.timeseries.bucket.BucketAttributes;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class ReportNodeTimeSeries implements Closeable {
```
cl-exense, line 23: Could/should this even be AutoCloseable? (sincere question, I don't really know if that has drawbacks/benefits).

step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/resolvedplan/ResolvedPlanBuilder.java
```diff
@@ -0,0 +1,89 @@
...
+    public ResolvedPlanNode buildResolvedPlan(Plan plan, Map<String, Object> bindings) {
+        return buildTreeRecursively(null, plan.getRoot(), null, objectPredicate, bindings);
+    }
+
+    private ResolvedPlanNode buildTreeRecursively(String parentId, AbstractArtefact artefactNode, String artefactPath, ObjectPredicate objectPredicate, Map<String, Object> bindings) {
+        String artefactId = artefactNode.getId().toString();
+        String artefactHash = artefactHashGenerator.generateArtefactHash(artefactPath, artefactId);
+
+        // Resolve children
+        List<AbstractArtefact> childrenArtefacts;
```
cl-exense, line 53: I'm torn: One could initialize childrenArtefacts/newArtefactPath with List.of()/null , in which case the first `else` and even `catch` implementation could go away/be empty. But I also feel that the current code, even if minimally redundant, is clearer in terms of what happens. I'll leave the decision to you, I'm fine with both.

cl-exense (changes requested):
Phew :) I admit that I'm still not fully grasping the purpose, but did my best to review it. In general it looks like good code, but there are a few hopefully somewhat useful comments and requests for (minor) changes.

step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/aggregated/ReportNodeTimeSeries.java
david-stephan: It is autocloesable (Closeable extends AutoCloseable) and can be used in a try with resource construct. However the time  serie is created and added to the Step controller context and available during the lifetime of  the controller. The Step context when closed will automatically close all stored object which are instance of and only of Closeable

step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/resolvedplan/ResolvedPlanBuilder.java
david-stephan: This code has been fully refectored in child branch 3131

david-stephan:
Merging this PR to master to be in sync with FE. Ticket SED-3350 created for follow up tasks based on PR review

PR #302, opened 2024-08-20
SED-3153 Update dependency management

cl-exense:


step-agent/pom.xml
```diff
@@ -51,7 +49,7 @@
 								<artifactItem>
 									<groupId>ch.exense.step</groupId>
 									<artifactId>step-grid-agent</artifactId>
-									<version>${step-grid.version}</version>
+<!--									<version>${step-grid.version}</version>-->
```
david-stephan, line 52: I assume this is not required anymore or the build would not go trough, the comment should be removed in that case.

step-plans/step-plans-core/pom.xml
```diff
@@ -47,19 +47,17 @@
...
-			<version>${step-api.version}</version>
 		</dependency>
 		<dependency>
 			<groupId>org.apache.commons</groupId>
 			<artifactId>commons-lang3</artifactId>
-			<version>${commons-lang.version}</version>
 		</dependency>
 
 		<!-- Required in compile scope as some classes of main code are using junit -->
+		<!-- FIXME: IS THIS A GOOD IDEA ???? -->
```
david-stephan, line 57: The "main" class PlanRunnerResultAssert is using junit.framework.Assert. This class is only used then in test classes, I'm not sure if it can be use in productive code (like Step junit runner....). To be checked with @jeromecomte

david-stephan (dismissed):

PR #325, opened 2024-10-06
SED-3437: validate Step Controller version in CLI

iegorov777:


step-cli/step-cli-launcher/src/main/java/step/cli/StepConsole.java
```diff
@@ -40,7 +44,7 @@
 
 @Command(name = "step",
         mixinStandardHelpOptions = true,
-        version = "step 1.0",
+        version = Constants.STEP_API_VERSION_STRING,
```
jeromecomte, line 47: We should define this centrally and use the same constant for the CLI and BE.

step-cli/step-cli-launcher/src/main/java/step/cli/StepConsole.java
```diff
@@ -209,11 +217,34 @@ private void checkApFolder(File param) throws IOException {
...
+                return new ControllerCredentials(stepUrl, authToken == null || authToken.isEmpty() ? null : authToken);
+            }
+
+            protected void checkStepControllerVersion() {
+                try {
+                    new ControllerVersionValidator(createControllerServicesClient()).validateVersions(getVersion());
+                } catch (ControllerVersionValidator.ValidationException e) {
+                    if (!force) {
+                        log.error("To execute the command please update you CLI application or use the " + FORCE + " option to ignore this validation");
+                        throw new StepCliExecutionException("CLI version mismatch detected", e);
```
jeromecomte, line 235: Not sure if this message would be clear for the user. Would clearly state that the version of the controller doesn't match the version of the CLI and also print the 2 values. Proposition: 

Error: Version Mismatch
The server version (vX.X.X) is incompatible with the current CLI version (vY.Y.Y).
Please ensure both the CLI and server are running compatible versions. {{if minor mismatch => "You can use the --force option to ignore this validation"}}

jeromecomte (changes requested):
Requesting minor changes

step-cli/step-cli-launcher/src/main/java/step/cli/StepConsole.java
iegorov777: @jeromecomte this constant is already in `step-core` (maybe not the best module to keep this, by the way) and it is reused in CLI and BE.

jeromecomte (commented):
Approving with a minor remark

step-cli/step-cli-launcher/src/main/java/step/cli/StepConsole.java
jeromecomte: Perfect. Just added a minor remark for the warning in case of minor mismatch

step-maven-plugin/src/main/java/step/plugins/maven/AbstractStepPluginMojo.java
```diff
@@ -110,6 +126,32 @@ protected ControllerCredentials getControllerCredentials() {
...
 
+	protected ControllerServicesClient createControllerServicesClient(){
+		return new ControllerServicesClient(getControllerCredentials());
+	}
+
+	protected void checkStepControllerVersion() throws MojoExecutionException {
+		try {
+			new ControllerVersionValidator(createControllerServicesClient()).validateVersions(getStepVersion());
+		} catch (ControllerVersionValidator.ValidationException e) {
+			String msg = "Version mismatch. The server version " + e.getResult().getServerVersion() + " is incompatible with the current maven plugin version " + e.getResult().getClientVersion() + ". Please ensure both the CLI and server are running compatible versions.";
```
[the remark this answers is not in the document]
iegorov777: @jeromecomte Agree. I've rewritten this warning message

jeromecomte (approved):

PR #330, opened 2024-10-14
SED-3435-added attributes to TS collections

lucianbaciu:


step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesControllerPlugin.java
```diff
@@ -36,6 +36,7 @@
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
```
david-stephan, line 39: import is not used

david-stephan (changes requested):
minor change requested, good otherwise

david-stephan (approved):

david-stephan (approved):

PR #333, opened 2024-10-17
SED-3499 Plan table performance issue with many hidden plans

david-stephan:


step-plans/step-plans-core/src/main/java/step/core/plans/PlanAccessorImpl.java
```diff
@@ -29,6 +29,11 @@ public class PlanAccessorImpl extends AbstractAccessor<Plan> implements PlanAcce
 
 	public PlanAccessorImpl(Collection<Plan> collectionDriver) {
 		super(collectionDriver);
+		createIndexIfNeeded();
+	}
+
+	private void createIndexIfNeeded() {
+		createOrUpdateIndex("visible");
```
jeromecomte, line 36: At some point we should somehow use constants or add a comment to the corresponding field as it can easily happen that we rename the field and the index wouldn't work anymore

jeromecomte (approved):

PR #337, opened 2024-10-20
SED-3451-handle removed axes from framework

lucianbaciu:


step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesHandler.java
```diff
@@ -330,7 +329,24 @@ private TimeSeriesAPIResponse mapToApiResponse(FetchBucketsRequest request, Time
                             .build();
                 }
                 bucketResponses.add(bucketResponse);
-            });
+            }
+//            axis.forEach(index -> {
```
david-stephan, line 333: remove commented code

david-stephan (changes requested):

david-stephan (approved):

PR #346, opened 2024-11-05
Sed 3408 filtering parameters by value searches protected fields as plain text

david-stephan:


step-controller/ControllerServer.run.xml
```diff
@@ -3,7 +3,7 @@
     <option name="MAIN_CLASS_NAME" value="step.controller.ControllerServer" />
     <module name="step-controller-backend" />
     <option name="PROGRAM_PARAMETERS" value="-config=&quot;src/test/resources/step.properties&quot;" />
-    <option name="VM_PARAMETERS" value="-Dhttp.keepAlive=true -Dhttp.maxConnections=50 -Dlogback.configurationFile=$PROJECT_DIR$/../step-distribution/step-distribution-parent/step-distribution-controller/template-controller/bin/logback.xml" />
+    <option name="VM_PARAMETERS" value="-Dhttp.keepAlive=true -Dhttp.maxConnections=50 -Dlogback.configurationFile=$PROJECT_DIR$/../step-distribution/step-distribution-controller/template-controller/bin/logback.xml" />
```
cl-exense, line 6: This requires a specific project layout on the develeper side (i.e., it requires the step-distribution to be checked out in a sibling directory). I'd rather copy the file into this project.

step-controller/step-controller-base-plugins/src/main/java/step/plugins/parametermanager/ParameterManagerControllerPlugin.java
```diff
@@ -63,7 +70,11 @@ public void serverStart(GlobalContext context) {
 		context.put("ParameterAccessor", parameterAccessor);
 
 		context.get(TableRegistry.class).register(ENTITY_PARAMETERS, new Table<>(collection, "param-read", true)
-				.withResultItemEnricher(p -> ParameterServices.maskProtectedValue(p)));
+				.withResultItemEnricher(ParameterServices::maskProtectedValue)
```
cl-exense, line 73: Should this really be an enricher? Enrichers can be disabled by the user. To unconditionally apply the masking, you can use the transformer instead of enricher.

cl-exense (changes requested):

step-controller/ControllerServer.run.xml
david-stephan: This is a fix for the existing configuration currently referring to a path that doesn't exist anymore since the refactoring of the distribution project. It's not the role of this PR to change this behavior.

step-controller/step-controller-base-plugins/src/main/java/step/plugins/parametermanager/ParameterManagerControllerPlugin.java
david-stephan: @cl-exense well this is pretty bad, because this is the legacy implementation of result enricher and the parameter table. I imagine making the enricher optional is recent and the parameter table hasn't been updated accordingly.

cl-exense: No, the enricher has always been "optional" in the sense that it can be disabled, as far as I know. That was the whole reason for introducing the transformer function which cannot be disabled.

PR #348, opened 2024-11-07
SED-3564 improve-time-selector

neogucky:
Solution for requesting specific time range from aggregated tree endpoint

### DIFF
--- a/step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/aggregated/AggregatedReportViewBuilder.java
+++ b/step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/aggregated/AggregatedReportViewBuilder.java
@@ -35,8 +35,14 @@ public class AggregatedReportViewBuilder {
         Objects.requireNonNull(request);
         Execution execution = executionAccessor.get(executionId);
         String aggregatedReportRoot = execution.getResolvedPlanRootNodeId();
-        ResolvedPlanNode resolvedPlanNode = resolvedPlanNodeAccessor.get(aggregatedReportRoot);
-        return recursivelyBuildAggregatedReportTree(resolvedPlanNode, request);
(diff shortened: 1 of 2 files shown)


step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/aggregated/AggregatedReportViewBuilder.java
```diff
@@ -35,8 +35,14 @@ public AggregatedReportView buildAggregatedReportView(AggregatedReportViewReques
         Objects.requireNonNull(request);
         Execution execution = executionAccessor.get(executionId);
         String aggregatedReportRoot = execution.getResolvedPlanRootNodeId();
-        ResolvedPlanNode resolvedPlanNode = resolvedPlanNodeAccessor.get(aggregatedReportRoot);
-        return recursivelyBuildAggregatedReportTree(resolvedPlanNode, request);
+        //Make sure the resolved Plan is available
```
cl-exense, line 38: Looks correct. Here's an equivalent solution that avoids explicit branches. I'll leave it up to you to decide which one you prefer and find more readable, I'd be fine with both.
```
        return Optional.ofNullable(execution.getResolvedPlanRootNodeId())
                .map(resolvedPlanNodeAccessor::get)
                .map(node -> recursivelyBuildAggregatedReportTree(node, request))
                .orElse(null);
```

cl-exense (approved):
Approving with optional alternative suggestion

step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/aggregated/AggregatedReportViewBuilder.java
david-stephan: feedback implemented

PR #349, opened 2024-11-11
SED-3571 initialization of lower resolutions incomplete

david-stephan:


jeromecomte (approved):

jeromecomte:
@david-stephan Just as heads-up: the version should be changed to an official release

PR #355, opened 2024-11-18
SED-3518 refactoring, support Json for input and output

david-stephan:


step-functions/step-functions-handler/src/main/java/step/functions/handler/ProcessBasedFunctionHandler.java
```diff
@@ -0,0 +1,268 @@
...
+    abstract protected String getProcessName();
+
+    @Override
+    public Output<JsonObject> handle(Input<JsonObject> input) throws Exception {
+        Map<String, String> properties = input.getProperties();
+        boolean debug = Boolean.parseBoolean(properties.getOrDefault(DEBUG, Boolean.FALSE.toString()));
+        boolean attachWorkFolder = Boolean.parseBoolean(properties.getOrDefault(ATTACH_WORK_FOLDER, Boolean.FALSE.toString()));
+        FileCommunicationType fileCommunicationType;
+        try {
+            fileCommunicationType = FileCommunicationType.valueOf(properties.getOrDefault(STEP_COMMUNICATION_FILE_TYPE, "JSON"));
```
jeromecomte, line 62: Just a detail but we could use the constant JSON instead of a string here

step-functions/step-functions-handler/src/main/java/step/functions/handler/ProcessBasedFunctionHandler.java
```diff
@@ -0,0 +1,268 @@
...
+        }
+    }
+
+    protected static void createProcessInputMessageAsJson(ManagedProcess process, Input<?> input) throws IOException {
+        //Create input properties file from keyword properties and keyword arguments
+        JsonObjectBuilder outputMessageBuilder = Json.createObjectBuilder();
+        outputMessageBuilder.add("inputs", (JsonObject)  input.getPayload());
+        JsonObjectBuilder propertiesJsonObject = Json.createObjectBuilder();
+        input.getProperties().forEach(propertiesJsonObject::add);
+        outputMessageBuilder.add("properties", propertiesJsonObject);
```
jeromecomte, line 162: Not sure but other attributes from the input object like the Keyword name or timeout might be useful too

step-functions/step-functions-handler/src/main/java/step/functions/handler/ProcessBasedFunctionHandler.java
```diff
@@ -0,0 +1,268 @@
...
+        OutputBuilder outputBuilder = new OutputBuilder();
+        File stepKeywordPropertiesFile = new File(process.getExecutionDirectory(), KEYWORD_OUTPUT_PROPERTIES_FILE);
+        if (stepKeywordPropertiesFile.exists() && stepKeywordPropertiesFile.isFile()) {
+            try (FileInputStream fis = new FileInputStream(stepKeywordPropertiesFile)) {
+                Properties outputProperties = new Properties();
+                outputProperties.load(fis);
+                outputProperties.keySet().stream().map(Object::toString).forEach(k -> outputBuilder.add(k,outputProperties.getProperty(k)));
+            } catch (IOException e) {
+                logger.error("Error while reading file " + KEYWORD_OUTPUT_PROPERTIES_FILE, e);
+                outputBuilder.appendError("Error while reading the Keyword output property file. See agent logs for more details.");
```
jeromecomte, line 184: We could also attach the exception to the output. Most users do not have access to the agent logs

step-functions/step-functions-handler/src/main/java/step/functions/handler/ProcessBasedFunctionHandler.java
```diff
@@ -0,0 +1,268 @@
...
+        Attachment attachment = AttachmentHelper.generateAttachmentFromByteArray(processOutput.getBytes(StandardCharsets.UTF_8), getSanitizedProcessName() + ".log");
+        outputBuilder.addAttachment(attachment);
+        String processErrorOutput = getProcessErrorOutput(process);
+        if (processErrorOutput != null && !processErrorOutput.isBlank()) {
+            Attachment errorAttachment = AttachmentHelper.generateAttachmentFromByteArray(processErrorOutput.getBytes(StandardCharsets.UTF_8), getSanitizedProcessName() + "Error.log");
+            outputBuilder.addAttachment(errorAttachment);
+        }
+    }
+
+    protected void attachProcessWorkFolder(ManagedProcess sikuli, OutputBuilder outputBuilder) {
```
jeromecomte, line 230: Wrong reference to sikuli

step-functions/step-functions-handler/src/main/java/step/functions/handler/json/JsonObjectDeserializer.java
```diff
@@ -0,0 +1,39 @@
...
+            return reader.readObject();
+        }
+    }
+
+    public static synchronized ObjectMapper getObjectMapper() {
+        if (objectMapper == null) {
+            objectMapper = new ObjectMapper();
+            // Register the custom deserializer for JsonObject
+            SimpleModule module = new SimpleModule();
+            module.addDeserializer(JsonObject.class, new JsonObjectDeserializer());
```
jeromecomte, line 34: Not sure but don't we have other modules to register for this layer?

jeromecomte (approved):
Approving with some remarks

step-functions/step-functions-handler/src/main/java/step/functions/handler/json/JsonObjectDeserializer.java
david-stephan: I don't see any that would be required. Most modules, (de)serializer that we register for other mapper are specific to Jackson, mongo, postgres, or our model classes... but I cannot be entirely sure either. Manual tests should help identify issues with that.

PR #360, opened 2024-12-03
SED-3471 complete-aggregated-report-during-execution-phase

david-stephan:


step-plans/step-plans-base-artefacts/src/main/java/step/planbuilder/BaseArtefacts.java
```diff
@@ -59,6 +59,16 @@ public static ForBlock for_(int start, int end) {
 		f.setDataSource(conf);
 		return f;
 	}
+
+	public static ForBlock for_(int start, int end, int threads) {
+		ForBlock f = new ForBlock();
```
jeromecomte, line 64: for_ could be refactored to use this method

step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/resolvedplan/ResolvedPlanBuilder.java
```diff
@@ -37,6 +44,7 @@ private ResolvedPlanNode buildTreeRecursively(String parentId, AbstractArtefact
         artefactClone.setChildren(null);
         ResolvedPlanNode resolvedPlanNode = new ResolvedPlanNode(artefactClone, artefactHash, parentId);
         resolvedPlanNodeAccessor.save(resolvedPlanNode);
+        resolvedPlanCache.put(resolvedPlanNode.artefactHash, resolvedPlanNode.getId().toHexString());
```
jeromecomte, line 47: To avoid high memory consumption, we should ensure that we cannot have situations we new hashes are generated for each node during long running executions with loops

step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/resolvedplan/ResolvedPlanBuilder.java
```diff
@@ -74,4 +82,36 @@ private ResolvedPlanNode buildTreeRecursively(String parentId, AbstractArtefact
...
+            });
+        }
+    }
+
+    private String findClosestParentNodeId(ReportNode parentReportNode, ReportNodeCache reportNodeCache) {
+        String planNodeId = resolvedPlanCache.get(parentReportNode.getArtefactHash());
+        if (planNodeId == null) {
+            ObjectId parentID = parentReportNode.getParentID();
+            if (parentID != null) {
+                planNodeId = findClosestParentNodeId(reportNodeCache.get(parentID), reportNodeCache);
```
jeromecomte, line 112: It shouldn't happen but we might throw a RuntimeException if we have to return null in this method i.e. if a node has no parent (root) and no corresponding planNodeId

jeromecomte (changes requested):
Requesting minor changes. Did think that it would be so simple :)

step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/resolvedplan/ResolvedPlanBuilder.java
david-stephan: As discussed we add here in the cache when building the resolved plan before execution. In the below method checkAndAddMissingResolvedPlanNode we exclude the work artefacts, the hash is generated in a correct way that should not generate unexpectedly different hash, so we should be safe

jeromecomte (approved):

PR #370, opened 2025-01-07
Sed 3338 cross execution dashboard

lucianbaciu:
@david-stephan here are the fixes. on top of the problems we discussed about, there is the .replaceAll("/r") which for me does not work, and it makes sense. If there is a reason for doing that which I miss, let me know. for me, both files are CRLF


step-plans/step-plans-core/src/main/java/step/core/artefacts/handlers/ArtefactHandler.java
```diff
@@ -22,6 +22,7 @@
 import org.bson.types.ObjectId;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import step.core.accessors.AbstractOrganizableObject;
```
david-stephan, line 25: Those new import are not used and not required

david-stephan (changes requested):
@lucianbaciu not sure why the new "replace" is required on your system but it doesn't hurt. Just cleanup the unused import, otherwise the PR is fine.

david-stephan (dismissed):

david-stephan (approved):

PR #379, opened 2025-02-07
SED-3721 attach single report nodes to aggregated report view

david-stephan:


step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/aggregated/AggregatedReportViewBuilder.java
```diff
@@ -59,7 +63,15 @@ private AggregatedReportView recursivelyBuildAggregatedReportTree(ResolvedPlanNo
...
-        return new AggregatedReportView(resolvedPlanNode.artefact, artefactHash, countByStatus, children, resolvedPlanNode.parentSource);
+        ReportNode singleInstanceReportNode = null;
+        if (countByStatus.values().stream().reduce(0L, Long::sum) == 1) {
+            singleInstanceReportNode = getSingleReportNodeInstance(executionId, artefactHash, request.range);
+        }
+        return new AggregatedReportView(resolvedPlanNode.artefact, artefactHash, countByStatus, children, resolvedPlanNode.parentSource, singleInstanceReportNode);
+    }
+
+    private ReportNode getSingleReportNodeInstance(String executionId, String artefactHash, ReportNodeTimeSeries.Range range) {
+        return reportNodeAccessor.getReportNodesByArtefactHash(executionId, artefactHash, (range != null) ? range.from : null, (range != null) ? range.to : null, 0, 1).findFirst().orElse(null);
```
jeromecomte, line 74: Not sure, but shouldn't we apply other filters (like status) too ?

jeromecomte, line 74: This query should return exactly 1 item. If it returns no item or more than 1, we might throw an unexpected exception as it would mean that the query doesn't match the distribution

jeromecomte (commented):
Requiring some clarifications

step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/aggregated/AggregatedReportViewBuilder.java
```diff
@@ -59,7 +63,15 @@ private AggregatedReportView recursivelyBuildAggregatedReportTree(ResolvedPlanNo
                 .collect(Collectors.toList());
         String artefactHash = resolvedPlanNode.artefactHash;
         Map<String, Long> countByStatus = reportNodesTimeSeries.queryByExecutionIdAndArtefactHash(executionId, artefactHash, request.range);
-        return new AggregatedReportView(resolvedPlanNode.artefact, artefactHash, countByStatus, children, resolvedPlanNode.parentSource);
+        ReportNode singleInstanceReportNode = null;
```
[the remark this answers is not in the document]
david-stephan: good point, I'll implement this and keep it enabled by default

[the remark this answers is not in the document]
jeromecomte: perfect. we should ensure that the FE is able to handle it properly

step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/aggregated/AggregatedReportViewBuilder.java
david-stephan: @jeromecomte I would not introduced an aggregation request filter on status as part or this ticket. Or did you mean to use the status of the single instance to resolve the report node for additional "safety". This could indeed be done.

david-stephan: @jeromecomte we could indeed let this function throw an exception. What would you do with it when building the report: rethrow up to the service (i.e. the FE won't be able to show the tree and display an error) or create a fake report node in which we set the error message?

jeromecomte: oh, you're right. I thought we had one already. If we don't, nothing has to be done, but in the future, if we would introduce one, we would have to think about it

jeromecomte (approved):

PR #384, opened 2025-02-14
SED-3738-generate-partial-aggregated-report

david-stephan:


step-plans/step-plans-base-artefacts/src/test/java/step/reporting/ResolvedPlanBuilderTest.java
```diff
@@ -68,14 +69,29 @@ public void simpleTest() throws IOException {
...
+        ReportNode reportNode = engine.getExecutionEngineContext().getReportNodeAccessor().getReportNodesByExecutionIDAndClass(result.getExecutionId(), "step.artefacts.reports.EchoReportNode").findFirst().orElseThrow(() -> new RuntimeException("No echo report node found"));
+        AggregatedReportViewBuilder.AggregatedReportViewRequest aggregatedReportViewRequest = new AggregatedReportViewBuilder.AggregatedReportViewRequest(null, true, reportNode.getId().toHexString(), false);
+        node = aggregatedReportViewBuilder.buildAggregatedReportView(aggregatedReportViewRequest);
+
+        logger.info("----------------------");
+        logger.info("Partial aggregated report tree");
+        logger.info("----------------------");
+        logger.info(node.toString());
+
+        assertEquals("ThreadGroup: 0x\n" +
```
jeromecomte, line 83: Instead of 0x (at least in the UI), I wouldn't display any count for parent nodes

step-plans/step-plans-base-artefacts/src/test/java/step/reporting/ResolvedPlanBuilderTest.java
```diff
@@ -306,6 +353,36 @@ public void planWithRecursiveCallPlan() throws IOException, InterruptedException
                         "            Set: 1x: PASSED > nextCount = 5\n" +
                         "            If: 1x: PASSED\n",
                 node.toString());
+
+        // Test partial aggregated tree for single Set
```
jeromecomte, line 357: Not sure to understand what we're exactly testing here. Do we test the case where the node doesn't belong to an "iteration"?

step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/aggregated/AggregatedReportViewBuilder.java
```diff
@@ -26,55 +30,125 @@ public class AggregatedReportViewBuilder {
...
         Execution execution = executionAccessor.get(executionId);
-        String aggregatedReportRoot = execution.getResolvedPlanRootNodeId();
         //Make sure the resolved Plan is available
-        return Optional.ofNullable(execution.getResolvedPlanRootNodeId())
-                .map(resolvedPlanNodeAccessor::get)
-                .map(node -> recursivelyBuildAggregatedReportTree(node, request))
-                .orElse(null);
+        ResolvedPlanNode rootResolvedPlanNode = Optional.ofNullable(execution.getResolvedPlanRootNodeId()).map(resolvedPlanNodeAccessor::get).orElse(null);
+        if (rootResolvedPlanNode == null) {
+            return null;
```
jeromecomte, line 56: Shouldn't we throw an exception here?

step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/aggregated/AggregatedReportViewBuilder.java
```diff
@@ -26,55 +30,125 @@ public class AggregatedReportViewBuilder {
...
+        } else if (request.selectedReportNodeId == null) {
+            // Generate complete aggregated report tree
+            return recursivelyBuildAggregatedReportTree(rootResolvedPlanNode, request, mainReportNodesTimeSeries, mainReportNodeAccessor, null);
+        } else {
+            // a node is selected to generate a partial aggregated report
+            try (ReportNodeTimeSeries localReportNodesTimeSeries = getInMemoryReportNodeTimeSeries()) {
+                InMemoryReportNodeAccessor inMemoryReportNodeAccessor = new InMemoryReportNodeAccessor();
+                Set<String> reportArtefactHashSet = buildPartialReportNodeTimeSeries(request.selectedReportNodeId, localReportNodesTimeSeries, inMemoryReportNodeAccessor);
+                // Only pass the reportArtefactHashSet if aggregate view filtering is enabled
+                reportArtefactHashSet = (request.filterResolvedPlanNodes) ? reportArtefactHashSet : null;
```
jeromecomte, line 66: What is the exact meaning of the switch filterResolvedPlanNodes?

jeromecomte (commented):
PR looks good. Just having a few minor questions

step-plans/step-plans-base-artefacts/src/test/java/step/reporting/ResolvedPlanBuilderTest.java
david-stephan: The UI does not display 0x, this is only the toString implementation. I find it more clear for the toString to explicitly write 0x, but le me know otherwise.

jeromecomte: I think that "0x" is a bit misleading here, as the parent was actually executed in this context but it is a detail and I wouldn't change it here if it only affects the toString

step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/aggregated/AggregatedReportViewBuilder.java
jeromecomte: Not sure if there was a reason for that. We can keep it as it is

jeromecomte (dismissed):

step-plans/step-plans-core/src/main/java/step/core/artefacts/handlers/ArtefactHandler.java
```diff
@@ -503,13 +506,27 @@ private REPORT_NODE createReportNode(ReportNode parentReportNode, ARTEFACT artef
 		node.setId(new ObjectId());
 		node.setName(getReportNodeName(artefact));
 		node.setParentID(parentReportNode.getId());
+		String parentPath = parentReportNode.getPath();
+		node.setPath(((parentPath != null) ? parentPath : "") + generateShortId(10));//Base64.getEncoder().encodeToString(node.getId().toByteArray()));
```
jeromecomte, line 510: - The performance of this part is critical and should be optimal. Calling generateShortId could generate an overhead. We should measure it under load and ensure that the implementation doesn't cause any contention
- Remove commented code
- object Id are 12byte large. Any reason for not using the id of the report node instead?

step-plans/step-plans-core/src/main/java/step/core/artefacts/handlers/ArtefactHandler.java
```diff
@@ -503,13 +506,27 @@ private REPORT_NODE createReportNode(ReportNode parentReportNode, ARTEFACT artef
...
 		node.setArtefactID(artefact.getId());
 		node.setExecutionID(context.getExecutionId().toString());
 		node.setStatus(ReportNodeStatus.NORUN);
 		node.setParentSource(parentSource);
 		return node;
 	}
 
+	public static String generateShortId(int length) {
+		String characters = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
```
jeromecomte, line 519: - Use constant. 
- Would move this method to dedicated class

step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/aggregated/AggregatedReportViewBuilder.java
```diff
@@ -26,55 +30,156 @@ public class AggregatedReportViewBuilder {
...
+            return null;
+        } else if (request.selectedReportNodeId == null) {
+            // Generate complete aggregated report tree
+            return new AggregatedReport(recursivelyBuildAggregatedReportTree(rootResolvedPlanNode, request, mainReportNodesTimeSeries, mainReportNodeAccessor, null));
+        } else {
+            // a node is selected to generate a partial aggregated report
+            try (ReportNodeTimeSeries localReportNodesTimeSeries = getInMemoryReportNodeTimeSeries()) {
+                InMemoryReportNodeAccessor inMemoryReportNodeAccessor = new InMemoryReportNodeAccessor();
+                AggregatedReport aggregatedReport = new AggregatedReport();
+                aggregatedReport.selectedReportNodeId = request.selectedReportNodeId;
```
jeromecomte, line 83: we're returning the same value as passed in the request. is it useful? shouldn't we return the id of the "iteration" reportNode ?

step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/aggregated/AggregatedReportViewBuilder.java
```diff
@@ -26,55 +30,156 @@ public class AggregatedReportViewBuilder {
...
+
+    private boolean isIterationNodeReport(ReportNode n) {
+        AbstractArtefact resolvedArtefact = n.getResolvedArtefact();
+        //All iteration work artefact are created with sequence artefact and names starting with "Iteration", we might change to a more robust implementation in the future
+        return resolvedArtefact != null && resolvedArtefact.isWorkArtefact() && resolvedArtefact.getAttribute(AbstractOrganizableObject.NAME).startsWith("Iteration");
+    }
+
+    private void ingestReportNodeRecursively(ReportNode reportNode, ReportNodeTimeSeries reportNodeTimeSeries, Map<String, ReportNode> singleReportNodes) {
+        reportNodeTimeSeries.ingestReportNode(reportNode);
+        singleReportNodes.put(reportNode.getArtefactHash(), reportNode);
```
jeromecomte, line 167: What happens if we have more than one reportnode per ArtefactHash (a loop in the sub tree)?
I assume that in this case the reportnode aren't used as we have more than one instance, but just to be sure

step-plans/step-plans-core/src/main/java/step/core/artefacts/handlers/ArtefactHandler.java
david-stephan: ObjectID is 24 chars. I started using simply these IDs for the path which works fine, but my concern was about the size of the records and index, so I first tried with a conversion to b64 (16chars) and finally to a custom random id of 10 chars. Honestly I don't know what is costing the most in term of performance, the size of the path or the conversion. Thinking about it I think the contention and uniqueness risk introduce by the random generator is not worth it, so I would revert to using the objectId directly. DBs usually do compression themselves so I'm not even sure that the b64 encoding brings anything.

step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/aggregated/AggregatedReportViewBuilder.java
david-stephan: not really useful indeed, even the iteration object id could not be really used by the FE, so I'll remove it altogether

step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/aggregated/AggregatedReportViewBuilder.java
david-stephan: yes report nodes are only used for single instances. This is implemented this way for memory optimization.

step-plans/step-plans-core/src/main/java/step/core/artefacts/handlers/ArtefactHandler.java
jeromecomte: Agree. We should anyway validate this during the load tests

jeromecomte (dismissed):

PR #386, opened 2025-02-17
SED-3705-dynamic-expressions-of-skipped-nodes-are-evaluated

david-stephan:


step-core/src/main/java/step/core/dynamicbeans/DynamicBeanResolver.java
```diff
@@ -92,6 +93,21 @@ public void evaluate(Object o, Map<String, Object> bindings) {
 		}
 	}
 
+	public void evaluateMandatoryFieldsEvenForSkippedArtefact(AbstractArtefact artefact, Map<String, Object> bindings) {
```
jeromecomte, line 96: The DynamicBeanResolver is not specific to AbstractArtefacts and applies to objects in general. For this reason I wouldn't add this specific method here. It would be nicer to make evaluateDynamicValue public and declare this method in a more specific class (like ArtefactHandler for instance=

step-plans/step-plans-core/src/main/java/step/core/artefacts/handlers/ArtefactHandler.java
```diff
@@ -114,11 +114,12 @@ public void createReportSkeleton(ReportNode parentReportNode, ARTEFACT artefact,
 		}
 
 		try {
-			dynamicBeanResolver.evaluate(artefact, getBindings());
+			dynamicBeanResolver.evaluateMandatoryFieldsEvenForSkippedArtefact(artefact, getBindings());
```
jeromecomte, line 117: If the artefact is not skipped, the method getBindings() and the evaluation of DynamicName and SkipNode are now evaluated twice. To avoid this, I would move this line to the line 120

step-plans/step-plans-core/src/main/java/step/core/artefacts/handlers/ArtefactHandler.java
```diff
@@ -177,16 +178,17 @@ public ReportNode execute(REPORT_NODE parentReportNode, ARTEFACT artefact, Map<S
...
```
jeromecomte, line 181: Idem

jeromecomte (approved):

PR #395, opened 2025-02-20
SED-3737 Consider execution_threads_auto parameter in agent auto-provisioning

cl-exense:
... plus some minor cleanup/refactoring

### DIFF
--- a/step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/handlers/functions/TokenForecastingExecutionPlugin.java
+++ b/step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/handlers/functions/TokenForecastingExecutionPlugin.java
@@ -27,6 +27,7 @@ import step.core.artefacts.reports.ReportNode;
 import step.core.execution.ExecutionContext;
 import step.core.execution.ExecutionEngineContext;
 import step.core.plugins.Plugin;
+import step.core.plugins.exceptions.PluginCriticalException;
 import step.engine.plugins.AbstractExecutionEnginePlugin;
 import step.threadpool.ThreadPool;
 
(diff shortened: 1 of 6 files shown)


cl-exense:
The actual change is pretty straightforward, but please double-check that this is indeed the correct place and behavior.

step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/handlers/functions/TokenForecastingExecutionPlugin.java
```diff
@@ -63,6 +64,14 @@ public void executionStart(ExecutionContext context) {
 
         // Token forecasting is always calculated, even if the agent provisioning is disabled
         TokenForecastingContext tokenForecastingContext = new TokenForecastingContext(availableAgentPools);
+        String executionThreadsAuto = context.getVariablesManager().getVariableAsString(ThreadPool.EXECUTION_THREADS_AUTO, null);
+        if (executionThreadsAuto != null && !executionThreadsAuto.isEmpty()) {
+            try {
+                tokenForecastingContext.setMaxParallelism(Integer.parseInt(executionThreadsAuto));
+            } catch (NumberFormatException e) {
+                logger.warn("Ignoring unparseable value for parameter {}: {}", ThreadPool.EXECUTION_THREADS_AUTO, executionThreadsAuto, e);
```
jeromecomte, line 72: Not sure if it wouldn't be better to throw an error in this case

jeromecomte (commented):
Having one minor comment

step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/handlers/functions/TokenForecastingExecutionPlugin.java
jeromecomte: Get it. In that case it would be cleaner to validate this value at the beginning of the execution and report an execution lifecycle error. This could be done in another ticket too. So nothing to be changed here. I would however log it as ERROR

cl-exense: Throwing error as requested

jeromecomte (approved):

PR #402, opened 2025-02-28
SED-3772-pass-properties-to-k-6-via-environment-variables

jeromecomte:


cl-exense (changes requested):
As discussed in the ticket, I'm really hesitant to use this kind of "dependency override" mechanism outside of local development and strongly suggest NOT to do it.

cl-exense (commented):
As discussed in the ticket, I'm fine with this, as long as it's a temporary measure and does not end up in the released product, so I would simply request to add a comment like

or so.

PR #405, opened 2025-03-03
SED-3781 migration-of-before-and-after-do-not-migrate-versioned-entities

david-stephan:


cl-exense (commented):
Code logic looks good, I’m just wondering why in some case we’re using `entity._entityClass` and in other cases `type` to identify the correct nodes (and for the unversioned functions it has been changed from type to entityClass, while for the versioned ones it "remains" type). That's somewhat confusing, so if this is indeed the correct behavior, then maybe a comment would help understand exactly what the differences are and why they exist.

david-stephan:
@cl-exense thanks for the review. There was indeed an error introduced mixing type property (used for entities) and _entityClass (used for entity versions). I pushed a fix for that. As of why we chose to use 2 different property names when we implemented the versioning, I must say I cannot remember the reason :-).

cl-exense (approved):

PR #406, opened 2025-03-05
SED-3367 Add 'involved agents' to execution

cl-exense:
This is conceptually pretty simple actually - after an execution has ended, just go through the execution report and gather all agent URLs.

This is a draft for now, as 1) it only is able to retrieve agent URLs from KW calls (but I think this actually is the only node type which explicitly provides that information), and 2) the code is slightly ugly because it needs to work by reflection as it doesn't have access to the actual class definitions (in terms of encapsulation: the module only knows about the supertype, but not the implementing subtypes). I did not find another solution that wouldn't be massively more complicated, but I'm open to suggestions (potentially a dedicated plugin just for that?)

### DIFF
--- a/step-plans/step-plans-core/src/main/java/step/core/execution/model/Execution.java
+++ b/step-plans/step-plans-core/src/main/java/step/core/execution/model/Execution.java
@@ -53,6 +53,7 @@ public class Execution extends AbstractOrganizableObject implements EnricheableO
 	private ExecutionParameters executionParameters;
 	private ExecutiontTaskParameters executiontTaskParameters;
 	private String resolvedPlanRootNodeId;
+	private String agentsInvolved;
 
 	public Execution() {
 		super();
@@ -228,6 +229,20 @@ public class Execution extends AbstractOrganizableObject implements EnricheableO
 
 	public void setResolvedPlanRootNodeId(String resolvedPlanRootNodeId) {
 		this.resolvedPlanRootNodeId = resolvedPlanRootNodeId;
+	}
+
+	/**
+	 * Returns a list of agents (i.e., their agent URLs) that were involved in the execution.
+	 * This is only populated after an execution has ended, and may not be available for all executions,
+	 * so null or empty values can be expected and should be treated as "unavailable/not applicable"
+	 * @return the list of involved agents (space-separated)
+	 */
+	public String getAgentsInvolved() {
+		return agentsInvolved;
+	}
+
+	public void setAgentsInvolved(String agentsInvolved) {
+		this.agentsInvolved = agentsInvolved;
 	}
 
 	@Override
--- a/step-plans/step-plans-core/src/main/java/step/engine/execution/ExecutionManager.java
+++ b/step-plans/step-plans-core/src/main/java/step/engine/execution/ExecutionManager.java
@@ -2,12 +2,17 @@ package step.engine.execution;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import step.core.artefacts.reports.ReportNode;
 import step.core.execution.ExecutionContext;
 import step.core.execution.model.Execution;
 import step.core.execution.model.ExecutionAccessor;
 import step.core.execution.model.ExecutionStatus;
 
+import java.lang.reflect.Method;
+import java.util.Objects;
 import java.util.function.Consumer;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public class ExecutionManager {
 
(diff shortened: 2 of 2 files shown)


david-stephan:
@cl-exense I haven't reviewed the PR in details, but I think we could use the FunctionPlugin and the hooks afterFunctionExecution and afterExecutionEnd to first populate the list of agents and persits it after the test. Going though all execution reports it not an option as for load test it can be an extremely  huge volume of data. I would also not update the execution in DB after each KW call because that would be a performance issue too.

cl-exense:
OK, second attempt, implementation is now better understandable and in fact almost trivial. I decided to make this a "first-class citizen" by introducing its own class; populating the values should now be essentially overhead-free and is done exactly "on use".

If you're fine conceptually, I'll add some small unit tests.

step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/handlers/CallFunctionHandler.java
```diff
@@ -210,10 +211,13 @@ protected void execute_(CallFunctionReportNode node, CallFunction testArtefact)
...
 				}
-				
-				node.setAgentUrl(token.getAgent().getAgentUrl());
+
+				String agentUrl = token.getAgent().getAgentUrl();
+				node.setAgentUrl(agentUrl);
 				node.setTokenId(token.getID());
-				
+
+				context.get(ExecutionAgentUrls.class).add(agentUrl);
```
jeromecomte, line 219: We're also adding local agentUrl here. Not sure if we shouldn't track only "remote" agents

step-plans/step-plans-core/src/main/java/step/core/execution/ExecutionContext.java
```diff
@@ -80,6 +81,7 @@ protected ExecutionContext(String executionId, ExecutionParameters executionPara
 		reportNode.setId(new ObjectId(executionId));
 		reportNodeCache.put(reportNode);
 		setCurrentReportNode(reportNode);
+		put(ExecutionAgentUrls.class, new ExecutionAgentUrls());
```
jeromecomte, line 84: The context map is made for plugins. Don't really see the advantage of not declaring this as a field in ExecutionContext

step-plans/step-plans-core/src/main/java/step/core/execution/model/ExecutionAgentUrls.java
```diff
@@ -0,0 +1,39 @@
...
+    // space-separated, and for convenience, sorted alphabetically (case-insensitive)
+    public String toString() {
+        return backingSet.stream()
+                .sorted(String.CASE_INSENSITIVE_ORDER)
+                .collect(Collectors.joining(" "));
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (o == null || getClass() != o.getClass()) return false;
```
jeromecomte, line 30: Where do we need this method?

jeromecomte (commented):
Requiring some clarifications

step-plans/step-plans-core/src/main/java/step/core/execution/ExecutionContext.java
cl-exense: Even better, if that's allowed it will make things even easier.

step-plans/step-plans-core/src/main/java/step/core/execution/model/ExecutionAgentUrls.java
cl-exense: That was just for good practice, using generate hashCode and equals to follow the usual java contract. Will probably be gone in next iteration.

step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/handlers/CallFunctionHandler.java
cl-exense: I'm not sure if that is the case. Do you have an example? Note that the URL is sometimes null, in which case it's simply ignored -- I think this is what the "local" agent is?

cl-exense: Well, it's implicitly checked in the add method anyway (which ignores null values), but I can also make it explicit.

step-plans/step-plans-core/src/main/java/step/core/execution/ExecutionContext.java
```diff
@@ -225,4 +230,20 @@ public Set<ExecutionVetoer> getExecutionVetoers() {
 	public void addExecutionVetoer(ExecutionVetoer executionVetoer) {
 		this.executionVetoers.add(executionVetoer);
 	}
+
+	public void addAgentUrl(String agentUrl) {
+		if (agentUrl != null) {
```
jeromecomte, line 235: AgentUrls is declared as final and should never be null

cl-exense: Nope, that's the entry which is to be added to the set :) . In principle, it *should* never be null if only non-local agents are added, but better safe than sorry, because otherwise it might throw an NPE.

jeromecomte: My bad, you're right. Was not completely awake :)

PR #420, opened 2025-03-26
SED-3854-timeseries-index-optimizations

cl-exense:


step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesControllerPlugin.java
```diff
@@ -126,6 +126,14 @@ public ExecutionEnginePlugin getExecutionEnginePlugin() {
 	public void initializeData(GlobalContext context) throws Exception {
 		super.initializeData(context);
 		timeSeries.createIndexes(new LinkedHashSet<>(List.of(new IndexField("eId", Order.ASC, String.class))));
+		timeSeries.getCollections().forEach(collection -> {
```
david-stephan, line 129: We could have introduced an equivalent method to createIndexes for the compound indexes directly in class TimeSeries, but that's a detail

david-stephan (dismissed):

PR #425, opened 2025-04-05
SED-2871: entity listing for AP

iegorov777:


step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/AutomationPackageManager.java
```diff
@@ -61,6 +61,10 @@
 
 public class AutomationPackageManager {
 
+    // TODO: maybe extract and reuse constants from YamlAutomationPackageSchemaGenerator
```
david-stephan, line 64: @iegorov777 indeed I would try to align this rather than redefining constants, let me know if it's a complex or high effort.

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/AutomationPackageManager.java
```diff
@@ -459,6 +463,24 @@ public AutomationPackageUpdateResult createOrUpdateAutomationPackage(boolean all
         }
     }
 
+    public Map<String, List<? extends Object>> getAllEntities(ObjectId automationPackageId) {
```
david-stephan, line 466: I would assume the entities are always extending AbstractOrganizableObject so I would be more specific for the returned type here

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/scheduler/AutomationPackageSchedulerHook.java
```diff
@@ -101,6 +101,11 @@ public void onPrepareStaging(String fieldName, AutomationPackageContext apContex
         }
     }
 
+    @Override
+    public Map<String, List<?>> getEntitiesForAutomationPackage(ObjectId automationPackageId, AutomationPackageContext automationPackageContext) {
```
david-stephan, line 105: I would assume the entities are always extending AbstractOrganizableObject so I would be more specific for the returned type here

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/AutomationPackageServices.java
```diff
@@ -251,4 +251,15 @@ public String getAutomationPackageDescriptorSchema() {
         return automationPackageManager.getDescriptorJsonSchema();
     }
 
+    @GET
+    @Path("{id}/entities")
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response listEntities(@PathParam("id") String id){
```
david-stephan, line 257: We usually define the concrete return type here, so map of String to List of entities

david-stephan (changes requested):

david-stephan (dismissed):

david-stephan (approved):

PR #426, opened 2025-04-07
Sed 3887 after sequence in plain text cause import errors

david-stephan:


step-plans/step-plans-parser/src/main/java/step/repositories/parser/ParsingContext.java
```diff
@@ -100,6 +104,9 @@ protected StackEntry pop() {
 	
 	public AbstractArtefact popCurrentArtefact() {
```
jeromecomte, line 105: To be fully consistant we should rename this popNonWrappingArtifact too or add the information as javadoc...

step-plans/step-plans-parser/src/test/java/step/repositories/parser/DefaultAndCustomStepParserTest.java
```diff
@@ -0,0 +1,83 @@
...
+import step.plans.nl.parser.PlanParser;
+
+public class DefaultAndCustomStepParserTest {
+
+
+
+    @Test
+    public void testComplexPlans() throws StepsParser.ParsingException {
+
+        StringBuilder sb = new StringBuilder();
```
jeromecomte, line 21: Using a resource would be easier to read

jeromecomte (approved):
Approving with 2 minimal remarks

PR #428, opened 2025-04-09
SED-3893 Fix forecasting handlers and execution_threads_auto support

cl-exense:


step-plans/step-plans-core/src/main/java/step/threadpool/ThreadPool.java
```diff
@@ -227,14 +227,37 @@ public Integer getAutoNumberOfThreads() {
...
+		executionContext.getVariablesManager().putVariable(executionContext.getCurrentReportNode(), EXECUTION_THREADS_AUTO_CONSUMED, Boolean.TRUE.toString());
 	}
-	
+
+	public Integer forecastNumberOfThreads(Integer originalNumberOfThreads) {
+		return forecastNumberOfThreads(originalNumberOfThreads, false);
+	}
+
+	public Integer forecastNumberOfThreads(Integer originalNumberOfThreads, boolean originalAsMaximum) {
+		Integer execution_threads_auto = getAutoNumberOfThreads();
```
jeromecomte, line 243: Why using "_" in the variable name? seems quite uncommon in java...

jeromecomte (approved):

step-plans/step-plans-core/src/main/java/step/threadpool/ThreadPool.java
cl-exense: Ehehe... yes, I know. This name directly corresponds to the variable name that the user sets, so I reused it in code to (potentially, if one knows the logic) highlight the relationship. And I just couldn't think of a better one :D

PR #434, opened 2025-04-16
SED-3472 execution-list-sorting-by-status-column-is-incorrect

david-stephan:


cl-exense (dismissed):
I assume this will need a new version but yeah, approving.

PR #438, opened 2025-04-28
SED-3958 Automation package race condition cause diverse executions e…

david-stephan:
…rrors

### DIFF
--- a/step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/AutomationPackageExecutionPlugin.java
+++ b/step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/AutomationPackageExecutionPlugin.java
@@ -4,22 +4,26 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import step.core.accessors.AbstractOrganizableObject;
 import step.core.execution.ExecutionContext;
+import step.core.execution.ExecutionEngineContext;
+import step.core.execution.model.ExecutionParameters;
 import step.core.plans.Plan;
 import step.core.plugins.IgnoreDuringAutoDiscovery;
 import step.core.plugins.Plugin;
 import step.core.plugins.exceptions.PluginCriticalException;
+import step.core.repositories.RepositoryObjectReference;
 import step.engine.plugins.AbstractExecutionEnginePlugin;
+import step.engine.plugins.FunctionPlugin;
 
-import static step.automation.packages.AutomationPackageLocks.BYPASS_AUTOMATION_PACKAGE_LOCK_FOR_ID;
-import static step.automation.packages.AutomationPackagePlugin.AUTOMATION_PACKAGE_READ_LOCK_TIMEOUT_SECS;
-import static step.automation.packages.AutomationPackagePlugin.AUTOMATION_PACKAGE_READ_LOCK_TIMEOUT_SECS_DEFAULT;
+import static step.automation.packages.AutomationPackageLocks.*;
+import static step.repositories.LocalRepository.getPlanId;
 
 /**
  * This execution plugin is used to prevent the update of automation packages currently in use by execution
- * Read lock on the package is started at execution start time and released when execution ends
+ * Read lock on the package is started at when initializing the execution and released when execution ends in finalizeExecutionContext
+ * It must run before FunctionPlugin.initializeExecutionContext which create the cached accessor for keywords
  */
 @IgnoreDuringAutoDiscovery
-@Plugin
+@Plugin(runsBefore= FunctionPlugin.class)
 public class AutomationPackageExecutionPlugin extends AbstractExecutionEnginePlugin {
 
     private static final Logger logger = LoggerFactory.getLogger(AutomationPackageExecutionPlugin.class);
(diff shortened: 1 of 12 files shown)


step-controller/step-controller-server/src/main/java/step/core/scheduler/ExecutionJob.java
```diff
@@ -18,25 +18,87 @@
...
+						throw new JobExecutionException("The execution task parameters for the provided schedule ID '" + executionTaskID + "' doesn't exists in the database anymore. This error may happen for race condition when deleting a schedule right when its job is triggered.");
+					}
+				}
+
+				ControllerSetting schedulerUsernameSetting = controllerSettingAccessor.getSettingByKey("scheduler_execution_username");
+				if (schedulerUsernameSetting != null) {
+					String schedulerUsername = schedulerUsernameSetting.getValue();
+					if (schedulerUsername != null && schedulerUsername.trim().length() > 0) {
+						// Override the execution user if the setting scheduler_execution_username is set
+						executiontTaskParameters.getExecutionsParameters().setUserID(schedulerUsername);
```
cl-exense, line 89: Shouldn't the value be trimmed also before setting, instead of only in the check above?

If so, a simple solution would be
```
String schedulerUsername = Optional.ofNullable(schedulerUsernameSetting.getValue()).map(String::trim).orElse(null);
if (schedulerUsername != null && !schedulerUsername.isEmpty()) {
```

step-controller/step-controller-server/src/main/java/step/core/scheduler/ExecutionJob.java
```diff
@@ -18,25 +18,87 @@
...
+	private final ControllerSettingAccessor controllerSettingAccessor;
+	private final AutomationPackageLocks automationPackageLocks;
 	private final String executionId;
-	
-	public ExecutionJob(ExecutionEngine executionEngine, String executionId) {
-		super();
+	private final String executionTaskID;
+
+	public ExecutionJob(ExecutionEngine executionEngine, ExecutionTaskAccessor executionTaskAccessor, ControllerSettingAccessor controllerSettingAccessor, AutomationPackageLocks lock, String executionId, String executionTaskID) {
 		this.executionEngine = executionEngine;
```
cl-exense, line 45: You may want to add `Objects.requireNonNull` around the assignments of parameters which are expected to be non-null. This way, should there ever come in a "wrongly null" value, an exception is thrown early instead of "somewhere else" at runtime. It also serves as a sort of "in-code" documentation concerning what can be expected to be null, and what not.

cl-exense (changes requested):
Did my best to review the changes as it's not exactly trivial to grasp everything at once, but in general it looks sensible, so there are basically just minor change requests.

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/AutomationPackageExecutionPlugin.java
```diff
@@ -31,74 +35,110 @@ public AutomationPackageExecutionPlugin(AutomationPackageLocks automationPackage
...
+        super.finalizeExecutionContext(executionEngineContext, executionContext);
+        AutomationPackageExecutionContext automationPackageExecutionContext = new AutomationPackageExecutionContext(executionContext);
+        Object lockId = executionContext.get(EXECUTION_CONTEXT_LOCK_ID);
+        if (lockId instanceof String) {
+            String automationPackageId = (String) lockId;
+            try {
+                automationPackageLocks.readUnlock(automationPackageId);
+                debugLog(automationPackageExecutionContext, "Released read lock on automation package.");
+            } catch (java.lang.IllegalMonitorStateException e) {
+                //occurs whenever the lock was not acquired in execution start due to timeout
```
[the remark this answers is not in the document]
david-stephan: Yes, that's correct, that's why I only added a debug log here (we still have to catch or throw the exception in the code to compile)

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/AutomationPackageExecutionPlugin.java
```diff
@@ -31,74 +35,110 @@ public AutomationPackageExecutionPlugin(AutomationPackageLocks automationPackage
...
             return false;
         } else {
-            Object byPassId = context.getExecutionParameters().getCustomField(BYPASS_AUTOMATION_PACKAGE_LOCK_FOR_ID);
+            Object byPassId = automationPackageExecutionContext.context.getExecutionParameters().getCustomField(BYPASS_AUTOMATION_PACKAGE_LOCK_FOR_ID);
             if (byPassId == null) {
                 return true;
             } else if (byPassId instanceof String) {
                 //Bypass lock if same automation package is already locked
-                return !automationPackageId.equals(byPassId);
+                return !automationPackageExecutionContext.automationPackageId.equals(byPassId);
```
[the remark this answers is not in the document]
david-stephan: well we explicitly check that it is not null in the if condition above, but I get the best practice and will adapt the code.

step-controller/step-controller-server/src/main/java/step/core/scheduler/ExecutionJob.java
david-stephan: it's legacy code and until now we were not trimming the user name, so I would not change the implementation and behavior.

cl-exense (approved):

PR #444, opened 2025-05-09
SED-3956-file-manager-file-names-containing-dir-are-considered-as-folder

david-stephan:
…folder


cl-exense:
Approving. (You've assigned the PR to me, so I'll let you review it again :-p ). Note that there's no strict need to bump the version here again after my last change to the step-grid PR, as that one only affects the unit tests. The version will get bumped again anyway before a release.

cl-exense (approved):

PR #456, opened 2025-06-10
SED-4014: fix overriding in AP fragments

iegorov777:


step-core/src/main/java/step/automation/packages/AutomationPackageHook.java
```diff
@@ -42,7 +50,15 @@ default void onPrepareStaging(String fieldName,
                                   AutomationPackage oldPackage,
                                   AutomationPackageStaging targetStaging) {
         // by default, we simply put the objects to staging
-        targetStaging.getAdditionalObjects().put(fieldName, objects);
+        if (objects != null) {
```
david-stephan, line 53: Same comment as above

david-stephan (dismissed):

PR #459, opened 2025-06-17
SED-2866-keyword-output-streaming

cl-exense:


step-controller-plugins/step-controller-plugins-streaming/src/main/java/step/plugins/streaming/ContextAwareWebsocketUploadEndpoint.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.List;
+import java.util.function.Consumer;
+
+public class ContextAwareWebsocketUploadEndpoint extends WebsocketUploadEndpoint {
+
+    public ContextAwareWebsocketUploadEndpoint(StreamingResourceManager manager, WebsocketServerEndpointSessionsHandler sessionsHandler) {
+        super(manager, sessionsHandler);
+    }
+
+    private String liveUploadContextId = null;
```
jeromecomte, line 25: Seeing that we store the resource metadata and references here, I'm asking me if we create one instance of this Endpoint for each uploaded resource. What is the lifecycle of this class?

step-controller-plugins/step-controller-plugins-streaming/pom.xml
jeromecomte, line 1: The streaming will be a core feature of the controller. I'm therefore wondering if there's a specific reason for declaring it in a dedicated module?

step-functions/step-functions-handler/src/main/java/step/functions/handler/liveupload/LiveUploadContextListeners.java
```diff
@@ -0,0 +1,76 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.lang.ref.ReferenceQueue;
+import java.lang.ref.WeakReference;
+import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Consumer;
+
+public class LiveUploadContextListeners {
```
jeromecomte, line 13: Not sure to see what makes this feature specific to Step. Couldn't it be moved to the streaming project?
If my understanding is correct, the idea is to pass a context id to the upload client and register for upload events related to this id. This feature seems not to be specific to Step

pom.xml
```diff
@@ -46,8 +46,10 @@
 		<maven.compiler.release>11</maven.compiler.release>
 		<!-- internal dependencies -->
 		<dependencies.version>2025.6.25</dependencies.version>
+
 		<step-grid.version>2025.6.27-685e89cd64523b01b02371a3</step-grid.version>
-		<step-framework.version>2025.6.26-685ce23164523b01b0137162</step-framework.version>
+		<step-framework.version>2025.7.2-6865213a343e7d5ee6f68a10</step-framework.version>
+		<step-streaming.version>0.0.0-SNAPSHOT</step-streaming.version>
```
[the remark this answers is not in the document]
cl-exense: Yes, I know, and we're almost there. It doesn't make sense just yet because there is nothing for the FE to see yet, but it's a matter of a few days.

step-controller-plugins/step-controller-plugins-streaming/src/main/java/step/plugins/streaming/StreamingResourceCollectionCatalogBackend.java
```diff
@@ -0,0 +1,45 @@
...
+import step.streaming.server.StreamingResourcesCatalogBackend;
+
+public class StreamingResourceCollectionCatalogBackend implements StreamingResourcesCatalogBackend {
+    private final StreamingResourceAccessor accessor;
+    public StreamingResourceCollectionCatalogBackend(GlobalContext context) {
+        accessor = new StreamingResourceAccessor(context.getCollectionFactory().getCollection(StreamingResource.COLLECTION_NAME, StreamingResource.class));
+    }
+    @Override
+    public String createResource(StreamingResourceMetadata metadata) {
+        StreamingResource entity = new StreamingResource();
```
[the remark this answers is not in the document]
cl-exense: Yes, and it's already foreseen. But currently there is no way to determine this information, and I don't know where it can come from. Maybe an additional parameter/header etc. on upload, we'll discuss that (along with the permission checks on download, and how they could be done).

step-controller-plugins/step-controller-plugins-streaming/src/main/java/step/plugins/streaming/StreamingResourcesControllerPlugin.java
```diff
@@ -0,0 +1,98 @@
...
+        String controllerUrl = context.getConfiguration().getProperty("controller.url");
+        if (controllerUrl == null) {
+            controllerUrl = "http://localhost:" + controllerPort;
+            logger.warn("controller.url is not set in step.properties, unable to determine correct URL for streaming services. Will use " + controllerUrl + " instead, but this will not be accessible remotely.");
+        }
+        // remove trailing slash if present
+        if (controllerUrl.endsWith("/")) {
+            controllerUrl = controllerUrl.substring(0, controllerUrl.length() - 1);
+        }
+        URI websocketBaseUri = URI.create(controllerUrl.replaceFirst("^http", "ws"));
```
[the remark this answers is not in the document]
cl-exense: Nothing :-) This was knowingly written this way because it transforms http:// into ws://, and https:// into wss://, which is exactly what we want.

step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/handlers/CallFunctionHandler.java
```diff
@@ -211,6 +214,13 @@ protected void execute_(CallFunctionReportNode node, CallFunction testArtefact)
 				node.setTokenId(token.getID());
 
 				Token gridToken = token.getToken();
+				LiveUploadContext uploadContext = LiveUploadContext.createNew();
+				LiveUploadContextListeners.registerListener(uploadContext.contextId, upload -> {
+					liveUploads.put(upload.reference.getUri(), upload);
```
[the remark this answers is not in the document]
cl-exense: That would be technically enough, yes. But it would only convey the information "there's something there", and would omit the information about what it is, and what state it's in (current size, status etc.). It's more efficient to directly add this information to the ReportNode IMO -- but this is exactly one of the parts that's still open.

step-controller-plugins/step-controller-plugins-streaming/pom.xml
cl-exense: We've discussed it with David and I put it along with the TimeSeries implementation, which is logically in a similar situation (core functionality, but implemented as plugin). I don't mind moving it elsewhere though.

step-functions/step-functions-handler/src/main/java/step/functions/handler/liveupload/LiveUploadContextListeners.java
cl-exense: There already is a similar, but even more lightweight concept, in the streaming project itself. There, it's really only about "has the status of resource X changed" events. This implementation plugs into that system but also goes beyond it by having Step-specific functionality  -- like "I'm interested about events (status changed, resource created etc.) in a particular Step call context. And it actually only looks scare because of the auto-cleanup logic :D

step-controller-plugins/step-controller-plugins-streaming/src/main/java/step/plugins/streaming/StreamingResourceCollectionCatalogBackend.java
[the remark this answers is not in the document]
jeromecomte: Perfect. I've sent you an invitation to discuss these open points

step-functions/step-functions-handler/src/main/java/step/functions/handler/liveupload/LiveUploadContextListeners.java
jeromecomte: I understand that the objects attached to the context are specific to Step but the logic enabling the creation of a context and its referencing everywhere it might be required is generic.

If we consider Step as a client of the streaming library the workflows could be the following:
1. The client creates a context in the streaming library and registers his own objects in this context
2. The client uses this context ID for resource uploads
3. Upon upload, if a context ID is set, the streaming library retrieves the context and passes it to the client everywhere it is required: StreamingResourcesCatalogBackend#createResource, StreamingResourcesStorageBackend#writeChunk, etc
This would for instance allow to:
- pass the ObjectEnricher to the StreamingResourcesCatalogBackend in order to enrich the resource
- pass the report node to any listener in order to add the id of the uploaded resources to the report node when an upload starts

PR #473, opened 2025-07-02
SED-3999-make execution range params optional

lucianbaciu:


lucianbaciu:
@david-stephan this is a change I did in order to be able to fetch the last execution, for a specific plan

david-stephan (approved):

neogucky (approved):

PR #481, opened 2025-07-15
SED-4100 Wrapping AP execution into test set has no default value for…

david-stephan:
… the number of threads

### DIFF
--- a/step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/execution/IsolatedAutomationPackageRepository.java
+++ b/step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/execution/IsolatedAutomationPackageRepository.java
@@ -136,7 +136,7 @@ public class IsolatedAutomationPackageRepository extends RepositoryWithAutomatio
 
     @Override
     protected boolean isWrapPlansIntoTestSet(Map<String, String> repositoryParameters) {
-        return Boolean.parseBoolean(repositoryParameters.getOrDefault(ArtifactRepositoryConstants.PARAM_WRAP_PLANS_INTO_TEST_SET, "true"));
+        return Boolean.parseBoolean(repositoryParameters.getOrDefault(ArtifactRepositoryConstants.PARAM_WRAP_PLANS_INTO_TEST_SET, "false"));
     }
 
     protected void updateLastExecution(Resource resource) {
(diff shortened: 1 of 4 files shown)


step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/execution/RepositoryWithAutomationPackageSupport.java
```diff
@@ -184,8 +184,13 @@ protected boolean isWrapPlansIntoTestSet(Map<String, String> repositoryParameter
 
     private Plan wrapAllPlansFromApToTestSet(PackageExecutionContext ctx, Map<String, String> repositoryParameters) {
         PlanBuilder planBuilder = PlanBuilder.create();
-        int numberOfThreads = Integer.parseInt(repositoryParameters.getOrDefault(ArtifactRepositoryConstants.PARAM_THREAD_NUMBER, "0"));
```
iegorov777, line 187: Is there some difference between `new TestSet()` (which uses 1 by default) and `new TestSet(0)` (which also leads to 1 via the hidden functionality)?

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/execution/IsolatedAutomationPackageRepository.java
```diff
@@ -136,7 +136,7 @@ public AutomationPackageFile restoreApFile(String contextId, Map<String, String>
 
     @Override
     protected boolean isWrapPlansIntoTestSet(Map<String, String> repositoryParameters) {
-        return Boolean.parseBoolean(repositoryParameters.getOrDefault(ArtifactRepositoryConstants.PARAM_WRAP_PLANS_INTO_TEST_SET, "true"));
```
iegorov777, line 139: @david-stephan 
Do you remember the reason why we used `true` by default in some (but not in all) cases? I suppose, `true` was used as default before we added the AP support into this Repositories, so we wanted to keep the old behavior (i.e wrap plans by default).

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/execution/RepositoryWithAutomationPackageSupport.java
david-stephan: @iegorov777 the functionality in the TestSetHandler leads only to 1 if the property "tec.execution.threads" is not overridden. I think it's better to align the logic with the default value of 1 for the field threads. I also prefer to use the empty constructor to make sure the default behavior is aligned with a TestSet directly created in the UI.

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/execution/IsolatedAutomationPackageRepository.java
david-stephan: I think we initially had and still have the default to true for MavenArtifactRepository, in this case we only support execution wrapping into a test set. Here I'm changing the default for the isolated and local AP execution which should also match the default we anyway set in the CLI, mvn plugin and UI.

PR #487, opened 2025-07-25
SED-4115 Activation expressions are not cached

jeromecomte:


david-stephan (dismissed):
approving with minor remarks

PR #501, opened 2025-08-19
SED-4178 Support local streaming, small refactorings

cl-exense:


step-cli/step-cli-launcher/src/main/java/step/cli/StepConsole.java
```diff
@@ -360,6 +362,9 @@ public static class ApExecuteCommand extends AbstractApCommand implements Callab
             @Option(names = {"--reportDir"}, description = "The local folder to store generated execution reports", defaultValue = "reports")
             protected File reportDir;
 
+            @Option(names = {"--streamingUploadsDir"}, description = "The local folder to store streaming uploads, only relevant for local executions. If not set, uploads will be processed but discarded. If set, the respective directory must exist.")
+            protected File streamingUploadsDir;
```
jeromecomte, line 366: It should be possible to access streamed attachments like normal attachments within the local execution of automation packages.
No sure if it is already supported for normal attachments but ideally we should store the streamed attachments along side with normal attachments within the report folder

jeromecomte (changes requested):

PR #503, opened 2025-08-21
SED-4189 Add lifecycle handling for LiveReporting, postpone CLI changes

cl-exense:


step-functions/step-functions-handler/src/main/java/step/functions/handler/FunctionMessageHandler.java
```diff
@@ -120,6 +121,8 @@ public OutputMessage handle(AgentTokenWrapper token, InputMessage inputMessage)
 			Output<?> output = functionHandler.handle(input);
 			measurementsBuilder.stopMeasure(customMeasureData());
 
+			liveReporting.close();
```
jeromecomte, line 124: Shouldn't we do this in a finally'

cl-exense: I thought about adding this to the "outer" finally block, but:
1. We cannot move the livereporting instantiation outside of the try block, because there is not enough information to instantiate it there.
2. even if that was possible, it still doesn't seem like the correct place in terms of "timing". LiveReporting (at least for now) should only be available while the handling takes place, and be immediately discarded after the handling. This is why it sits so "tightly" around it.
3. I don't think there is a risk of it being missed, as the functionHandler.handle() itself is not throwing exceptions (otherwise all the following logic in this block would also not work correctly, which would have resulted in broken executions already).

Not sure if that was understandable, but basically I'm trying to say: I think this is correct as it is :)

PR #509, opened 2025-09-02
SED-4206 groovy pool expose more tuning options

david-stephan:


step-core/src/main/java/step/expressions/GroovyPool.java
```diff
@@ -42,8 +49,16 @@ public GroovyPool(GroovyPoolFactory groovyPoolFactory, int poolMaxTotal, int poo
 			pool.setMaxTotalPerKey(poolMaxTotalPerKey);
 			pool.setMaxIdlePerKey(poolMaxIdlePerKey);
 			pool.setBlockWhenExhausted(true);
-			pool.setTimeBetweenEvictionRunsMillis(30000);
-			pool.setMinEvictableIdleTimeMillis(-1);
+			pool.setTimeBetweenEvictionRuns(Duration.ofMillis(30000));;
+			pool.setMinEvictableIdle(Duration.ofMillis(-1));
+			if (monitoringIntervalSeconds != null) {
+				scheduler = Executors.newScheduledThreadPool(1, r -> {
```
jeromecomte, line 55: new ThreadFactoryBuilder()
					.setNameFormat("groovy-pool-monitor-%d")
					.setDaemon(false)
					.build() from guava would do this. 

The convention seems to use lower case names: groovy-pool-monitor

step-core/src/main/java/step/expressions/GroovyPool.java
```diff
@@ -82,9 +91,44 @@ public void returnShell(GroovyPoolEntry entry) {
 	}
 
 
+	private void checkPoolHealth() {
+		try {
+			Map<String, Integer> numActivePerKey = pool.getNumActivePerKey();
+			int active = numActivePerKey.values().stream().mapToInt(Integer::intValue).sum();
```
jeromecomte, line 97: Why not using getNumActive() directly?

step-core/src/main/java/step/expressions/GroovyPool.java
```diff
@@ -82,9 +91,44 @@ public void returnShell(GroovyPoolEntry entry) {
...
+			logger.debug("Error monitoring pool", e);
+		}
+	}
+
+	private void checkHotKeys(Map<String, Integer> numActivePerKey ) {
+		Map<String, Integer> numWaitersByKey = pool.getNumWaitersByKey();
+		int maxPerKey = pool.getMaxTotalPerKey();
+		numActivePerKey.forEach((key,activePerKey) -> {
+			if (activePerKey >= maxPerKey * 0.9) {
+				logger.warn("Pool exhausted for expression '{}': {}/{} active",
```
jeromecomte, line 114: Above we say "nearly exhausted" for this situation

jeromecomte, line 114: We might add some recommendations

step-core/src/main/java/step/expressions/GroovyPool.java
david-stephan: getNumActive is calculated with total max size - (sum of idle per key). Since we also  need the active per key later, using this approach is more effecient (you only need to broth the idle per key once.

step-core/src/main/java/step/expressions/GroovyPool.java
```diff
@@ -52,11 +54,11 @@ public GroovyPool(GroovyPoolFactory groovyPoolFactory, int poolMaxTotal, int poo
...
 			pool.setMinEvictableIdle(Duration.ofMillis(-1));
 			if (monitoringIntervalSeconds != null) {
-				scheduler = Executors.newScheduledThreadPool(1, r -> {
-					Thread thread = new Thread(r, "GroovyPoolMonitor");
-					thread.setDaemon(true);
-					return thread;
-				});
+				ThreadFactory threadFactory = new ThreadFactoryBuilder()
+						.setNameFormat("groovy-pool-monitor-%d")
+						.setDaemon(false)
```
jeromecomte, line 59: It was true before

PR #516, opened 2025-09-17
SED-4222 fix-performance-and-resource-issues-under-heavy-load

cl-exense:


step-functions/step-functions-handler/src/main/java/step/functions/handler/FunctionMessageHandler.java
```diff
@@ -181,45 +186,19 @@ private LiveReporting initializeLiveReporting(Map<String, String> properties) th
...
-				if (agentConfUrl != null) {
-					// just a sanity check really
-					if (!agentConfUrl.matches("^https?://.+")) {
-						throw new IllegalArgumentException("Invalid URL in 'step.reporting.url' (agent-side configuration): " + agentConfUrl);
+				URI websocketUploadUri = getWebsocketUploadUri(properties, uploadContextId);
+				@SuppressWarnings("unchecked") Class<StreamingUploadProvider> providerClass = (Class<StreamingUploadProvider>) Thread.currentThread().getContextClassLoader().loadClass("step.streaming.websocket.client.upload.WebsocketUploadProvider");
+				if (webSocketContainer == null) {
+					synchronized(FunctionMessageHandler.this) {
+						if (webSocketContainer == null) {
+							webSocketContainer = providerClass.getDeclaredMethod("instantiateWebSocketContainer").invoke(null);
```
jeromecomte, line 194: If instantiateWebSocketContainer would be part of the interface StreamingUploadProvider there would be no need for reflection

step-functions/step-functions-handler/src/main/java/step/functions/handler/FunctionMessageHandler.java
```diff
@@ -68,6 +69,8 @@ public class FunctionMessageHandler extends AbstractMessageHandler {
 	private final ObjectMapper mapper;
 
 	private ExecutorService liveReportingExecutor;
+	// This is actually a Jakarta WebsocketContainer, but instantiated dynamically using a separate class loader
+	private volatile Object webSocketContainer;
```
jeromecomte, line 73: This could be wrapped in a class of the streaming api in order not to work with Object

step-functions/step-functions-handler/src/main/java/step/functions/handler/FunctionMessageHandler.java
```diff
@@ -83,7 +86,7 @@ public void init(AgentTokenServices agentTokenServices) {
 		super.init(agentTokenServices);
 
 		String liveReportingPoolSizeAgentPropsKey = "step.reporting.livereporting.poolsize";
-		int liveReportingPoolSizeDefault = 100;
+		int liveReportingPoolSizeDefault = 100; // ThreadPools.getDefaultPoolSize();
```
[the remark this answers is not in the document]
cl-exense: Lowered to 10 by default

step-functions/step-functions-handler/src/main/java/step/functions/handler/FunctionMessageHandler.java
```diff
@@ -99,7 +102,9 @@ public void init(AgentTokenServices agentTokenServices) {
 						})
 						.orElse(liveReportingPoolSizeDefault);
 
-		liveReportingExecutor = Executors.newFixedThreadPool(liveReportingPoolSize, NamedThreadFactory.create("livereporting-executor"));
+		// FIXME: figure out the final strategy here
```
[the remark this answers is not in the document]
cl-exense: For now staying with .newFixedThreadPool

step-functions/step-functions-handler/src/main/java/step/functions/handler/FunctionMessageHandler.java
cl-exense: I don't currently see a good way of implementing this without overcomplicating things, but I'm open to suggestions.

step-functions/step-functions-handler/src/main/java/step/functions/handler/FunctionMessageHandler.java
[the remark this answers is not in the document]
jeromecomte: As discussed let's use a fixedThreadPool as strategy with bouded queue

step-functions/step-functions-handler/src/main/java/step/functions/handler/FunctionMessageHandler.java
cl-exense: I actually tried that, but then I got ClassCastExceptions all over the place because the classes loaded from the two classloaders were deemed incompatible, so I just left this as is.

PR #577, opened 2025-12-18
SED-4157-executions history

lucianbaciu:


step-plans/step-plans-base-artefacts/src/main/java/step/reporting/ExecutionHistoryReportPlugin.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    @Override
+    public void afterExecutionEnd(ExecutionContext context) {
+        ExecutionAccessor executionAccessor = context.getExecutionAccessor();
+        Execution execution = context.getExecutionManager().getExecution();
+        // endTime is not set here
+        long searchBeforeTimestamp = System.currentTimeMillis() - 1;
+        List<ExecutionResultSnapshot> pastExecutionsSnapshots = executionAccessor.getLastEndedExecutionsByCanonicalPlanName(execution.getCanonicalPlanName(), 10, searchBeforeTimestamp)
+                        .stream()
```
david-stephan: getLastEndedExecutionsByCanonicalPlanName should return a stream directly

step-plans/step-plans-core/src/main/java/step/core/execution/model/ExecutionAccessorImpl.java
```diff
@@ -174,4 +176,22 @@ public List<Execution> getLastEndedExecutionsBySchedulerTaskID(String schedulerT
 						order, 0, limit, 0)
 				.collect(Collectors.toList());
 	}
+
+	@Override
+	public List<Execution> getLastEndedExecutionsByCanonicalPlanName(String canonicalPlanName, int limit, Long from) {
```
david-stephan: The provided timestamp is used to filter execution ending before it, please rename accordingly

step-plans/step-plans-core/src/main/java/step/core/execution/ExecutionEngineRunner.java
```diff
@@ -86,9 +86,27 @@ protected PlanRunnerResult execute() {
 				addImportResultToExecution(importResult);
 				saveFailureReportWithResult(ReportNodeStatus.VETOED);
 			} else {
+				String canonicalPlanName;
 				try {
-					Plan plan = getPlanFromExecutionParametersOrImport();
+					ExecutionParameters executionParameters = executionContext.getExecutionParameters();
```
david-stephan: So as discussed please revert to calling the method getPlanFromExecutionParametersOrImport, and imply update the execution canonical name from within that method. We already persist the import result their.

step-plans/step-plans-core/src/main/java/step/core/repositories/AbstractRepository.java
```diff
@@ -39,6 +39,18 @@ private static Map<String, String> getCanonicalRepositoryParameters(Set<String>
                 .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue)) : null;
     }
 
+    public static String getCanonicalPlanName(Map<String, String> repositoryParameters) {
+        if (repositoryParameters == null || repositoryParameters.isEmpty()) {
+            return "";
```
david-stephan: We should return null to be consistent and let the caller know that no canonical name exist

step-plans/step-plans-core/src/main/java/step/core/repositories/AbstractRepository.java
```diff
@@ -39,6 +39,18 @@ private static Map<String, String> getCanonicalRepositoryParameters(Set<String>
                 .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue)) : null;
     }
 
+    public static String getCanonicalPlanName(Map<String, String> repositoryParameters) {
```
david-stephan: This cannot be a static method since you need access to the canonical keys set

david-stephan (changes requested):
@lucianbaciu in addition to the feedback from Gemini which are valid too, please see my comments

step-plans/step-plans-core/src/main/java/step/core/execution/model/ExecutionAccessorImpl.java
```diff
@@ -174,4 +177,21 @@ public List<Execution> getLastEndedExecutionsBySchedulerTaskID(String schedulerT
 						order, 0, limit, 0)
 				.collect(Collectors.toList());
 	}
+
+	@Override
+	public Stream<Execution> getLastEndedExecutionsByCanonicalPlanName(String canonicalPlanName, int limit, Long from) {
```
david-stephan: Please rename from to searchBeforeTimestamp for clarity and check my comment on the plugin class using this method.

step-plans/step-plans-core/src/main/java/step/core/execution/ExecutionEngineRunner.java
```diff
@@ -196,7 +196,10 @@ private Plan getPlanFromExecutionParametersOrImport() throws PlanImportException
 	}
 
 	private void addImportResultToExecution(ImportResult importResult) {
-		updateExecution(e -> e.setImportResult(importResult));
+		updateExecution(e -> {
+			e.setImportResult(importResult);
+			e.setCanonicalPlanName(importResult.getCanonicalPlanName());
```
david-stephan: since importResult is already part of the execution model,  we should not add an explicit canonicalPlanName to the execution class, we can always access the value with execution.importResult.canonicalPlanName directly. Doing this is error prone as one will always have to remember that these fields are duplicated and must be kept in sync.

david-stephan (changes requested):

step-plans/step-plans-base-artefacts/src/main/java/step/reporting/ExecutionHistoryReportPlugin.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Override
+    public void afterExecutionEnd(ExecutionContext context) {
+        Configuration configuration = context.getConfiguration();
+        Integer countItems = configuration.getPropertyAsInteger(EXECUTIONS_HISTORY_COLLECT_COUNT, 10);
+        ExecutionAccessor executionAccessor = context.getExecutionAccessor();
+        Execution execution = context.getExecutionManager().getExecution();
+        // endTime is not set here
+        long searchBeforeTimestamp = execution.getEndTime() != null ? execution.getEndTime() : System.currentTimeMillis() - 1;
+        List<ExecutionResultSnapshot> pastExecutionsSnapshots = executionAccessor.getLastEndedExecutionsByCanonicalPlanName(execution.getCanonicalPlanName(), countItems, searchBeforeTimestamp)
```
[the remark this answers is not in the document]
david-stephan: I would not try to play around with timestamps, we know the execution ID so we should have a filter explicitly excluding it in the query to the DB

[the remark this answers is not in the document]
lucianbaciu: alright

david-stephan (changes requested):

david-stephan (dismissed):

david-stephan (dismissed):

david-stephan (approved):

PR #615, opened 2026-03-19
SED-4539 automation package collection and accessor

cmisev:


step-core/src/main/java/step/automation/packages/AutomationPackageArchive.java
```diff
@@ -61,6 +61,8 @@ public String getAutomationPackageName() {
 
     abstract public boolean hasAutomationPackageDescriptor();
 
+    abstract public URL getDescriptorYamlUrl();
```
cl-exense: Heads up: this method is not implemented in the DotNetAutomationPackageArchive, it's therefore currently not possible to build an enterprise distribution for this branch.

Judging from the other methods, I guess the correct implementation there is
```java
    @Override
    public URL getDescriptorYamlUrl() {
        throw new UnsupportedOperationException(NOT_SUPPORTED_WITH_DOT_NET_DLL_AUTOMATION_PACKAGE);
    }

```

cl-exense (changes requested):
Looks mostly fine!

Don't be scared by the number of comments, this is actually a pretty good PR, and the requested changes are mostly simple.

Note that I have not understood all of the logic in deep detail, but the overall concept, and I've focused mostly on code understandability rather than implementation details.

cl-exense:
Closing (decling) as discussed; merged into parent SED-4429 branch.

PR #627, opened 2026-04-09
SED-4637 Bump version and simplify migration tasks

cl-exense:


cl-exense:
Both commits are trivial, the first one updates the version to 3.30.0, the second one removes all runDowngradeScript methods. Note this PR is against the larger Jetty 12 PR which is already approved and finished except for the missing migration task.

david-stephan (approved):

PR #633, opened 2026-04-15
SED-4574 analytics-introduce-counter-gauge-histogram-metric-types-with-live-reporting-and-keyword-output-support

david-stephan:


step-controller/step-controller-server/src/main/java/step/core/metrics/MetricsExecutionPlugin.java
```diff
@@ -361,5 +395,73 @@ private void enrichWithAdditionalAttributes(Measurement measurement, ExecutionCo
...
+            // 2. notify handler
+            for (SamplesHandler samplesHandler : SamplesExecutionPlugin.SAMPLES_HANDLERS) {
+                samplesHandler.afterExecutionEnd(context);
+            }
+            // 3. Cleanup
+            MetricHeartbeatRegistry.getInstance().removeExecution(context.getExecutionId());
+            // Clean up thread group counters for this execution.
+            // Prometheus label cleanup (with the 70s scrape-window delay) is handled by PrometheusHandler.
+            String prefix = context.getExecutionId() + "|";
+            threadGroupCounts.entrySet().removeIf(e -> e.getKey().startsWith(prefix));
```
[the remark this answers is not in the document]
david-stephan: threadGroupCounts is private final field initialized with a ConcurrentHashMap wich cover concurrent accesses and prevent it to be null

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesControllerPlugin.java
```diff
@@ -73,30 +81,37 @@ public void serverStart(GlobalContext context) {
...
+        Set<String> includedAttributes = Arrays.stream(configuration.getProperty(TIME_SERIES_ATTRIBUTES_PROPERTY, "").split(","))
+                .filter(s -> !s.isEmpty())
+                .collect(Collectors.toSet());
+        Set<String> excludedAttributes = Arrays.stream(configuration.getProperty(TIME_SERIES_EXCLUDED_ATTRIBUTES_PROPERTY, TIME_SERIES_EXCLUDED_ATTRIBUTES_DEFAULT).split(","))
+            .filter(s -> !s.isEmpty())
+            .collect(Collectors.toSet());
+
+        if (!includedAttributes.isEmpty() && !excludedAttributes.isEmpty()) {
+            throw new PluginCriticalException("Setting both the properties " + TIME_SERIES_ATTRIBUTES_PROPERTY + " and "  + TIME_SERIES_EXCLUDED_ATTRIBUTES_PROPERTY + " is not allowed.");
+        }
```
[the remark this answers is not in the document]
david-stephan: Nothing critical is executed before, but moved up nevertheless

step-controller-plugins/step-controller-plugins-metrics-raw/src/main/java/step/plugins/metrics/raw/MeasurementAccessor.java
cl-exense, line 1: Unless I got something wrong, this should probably be renamed to RawMetricAccessor to be inline with the rest of the refactorings/renamings.

The same goes for "measurement" occurrences in method names of this class.

step-controller-plugins/step-controller-plugins-metrics-raw/src/main/java/step/plugins/metrics/raw/RawSamplesServices.java
```diff
@@ -0,0 +1,92 @@
...
+import step.framework.server.security.Secured;
+import step.core.metrics.ExecutionMetricSample;
+
+import java.util.*;
+import java.util.stream.Stream;
+
+@Singleton
+@Path("raw-samples")
+@Tag(name = "RawMetricSamples")
+public class RawSamplesServices extends AbstractStepServices {
```
cl-exense: Proposed rename to RawMetricSample(s)Services

step-controller-plugins/step-controller-plugins-metrics-raw/src/main/java/step/plugins/metrics/raw/MetricSampleAccessor.java
```diff
@@ -0,0 +1,28 @@
...
+import step.core.collections.Collection;
+import step.core.collections.Filters;
+import step.core.collections.filters.Equals;
+import step.core.metrics.MetricsExecutionPlugin;
+import step.core.metrics.ExecutionMetricSample;
+
+import java.util.Map;
+import java.util.stream.Stream;
+
+public class MetricSampleAccessor extends AbstractAccessor<ExecutionMetricSample> {
```
cl-exense: Given the "target" type, this should be called ExecutionMetricSampleAccessor

step-controller-plugins/step-controller-plugins-metrics-raw/src/main/java/step/plugins/metrics/raw/RawMetricsControllerPlugin.java
```diff
@@ -0,0 +1,88 @@
...
+@Plugin
+public class RawMetricsControllerPlugin extends AbstractControllerPlugin {
+
+    private static final Logger logger = LoggerFactory.getLogger(RawMetricsControllerPlugin.class);
+
+    private Collection<Document> measurementsCollection;
+    private Collection<ExecutionMetricSample> metricSamplesCollection;
+
+    @Override
+    public void serverStart(GlobalContext context) throws Exception {
```
cl-exense, line 49: This method is a good example showcasing, by now, a LOT of confusing naming.

We have multiple "things" all orbiting around the same concepts, but inconsistently named. First off, we have metrics, raw metrics, (metric) samples, and raw (metric) samples. What are the differences?

Then, we have a class Measurement. It's just a HashMap, and it relates to timing. The timeseries plugin uses a Collection<Measurement>, so far, so good. But then there's a MeasurementAccessor which uses the same collection, except as a collection of Documents.

(Oh BTW, in the API, we have Measures, which get mapped to Measurements when they are received). But I don't know how they relate to metrics or raw metrics.

Then we have a class ExecutionMetricSample, but the collection is called metricSamples.

Then we have a method called register**Samples**Handler which takes a **RawMetricSamples**Handler which in turn requires a **measurement**Accessor (remember, that accesses **Documents**, not Measurements), and a **metrics**Accessor.

And finally we have RawSamplesServices.

:-) :-D :-O ;-(

Don't get me wrong, I'm not trying to bully anyone, but this is getting extremely difficult to mentally bring together.

As far as I understand,
- Metric and RawMetric are related but not identical. Methods or classes involving them should clearly and unambiguously involve either "Metric" or "RawMetric". It should also be made clear somewhere how they relate and why they are different, and why different classes are required.
- the same goes for the various samples. They should always be referred to as MetricSample or RawMetricSample in method or class names, to make it clear what's used.
- Measures/Measurements/ their collections/accessors should also at least be consistently named (and potentially consolidated). That probably means at least renaming the MeasurementAccessor to MeasurementDocumentAccessor to make this clearer.

cl-exense (changes requested):
I'm only 20% through, but will already leave some feedback as I'll have to stop here for the time being - please take a look at the comments, especially the last one. If that one reads a little frustrated, that's because it is ;-) .

No offense meant, I understand that this is all pretty complex and also organically grown, but IMO this is something that really should be addressed -- simply because it makes it unnecessarily complicated to navigate through and relate the various concepts.

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesControllerPlugin.java
```diff
@@ -174,6 +206,21 @@ public void initializeData(GlobalContext context) throws Exception {
         initTimeSeriesCollectionsData(asyncTaskManager);
     }
 
+    private void warnAboutDisabledCollections(TimeSeriesConfig config) {
```
cl-exense: Fine with the wording, I just wonder about the loglevel (IMO this is more an INFO), and if this really needs to be shown every single time. I personally think it's better to put this information in the documentation / knowledge base, something like "if you disable... you can delete the corresponding DB table/collection".

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesHandler.java
```diff
@@ -36,55 +36,86 @@
...
         this.asyncTaskManager = asyncTaskManager;
         this.timeSeries = timeSeries;
         this.samplingLimit = samplingLimit;
-        this.attributesWithPrefix = this.timeSeriesAttributes
+        this.includedAttributesWithPrefix = this.timeSeriesIncludedAttributes
             .stream()
             .map(x -> ATTRIBUTES_PREFIX + x)
-            .collect(Collectors.toList());
+            .collect(Collectors.toSet());
+        this.exclduedAttributesWithPrefix = this.timeSeriesExcludedAttributes
```
cl-exense: typo in field name

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesHandler.java
```diff
@@ -98,33 +129,52 @@ private TimeSeriesAPIResponse getTimeSeriesFromRawMeasurements(FetchBucketsReque
...
                 stream.forEach(measurement -> {
                     count.increment();
                     timeSeriesBucketingHandler.ingestExistingMeasurement(measurement);
                 });
             }
+            // Iterate over each metric samples and ingest it again if metricSampleCollection exists
+            if (metricSampleCollection != null) {
+                try (Stream<ExecutionMetricSample> stream = metricSampleCollection.findLazy(filterMetricSamples, metricsSearchOrder, null, null, 0)) {
+                    stream.forEach(metricSample -> {
+                        if (metricSample != null) {
```
cl-exense: An equivalent implementation would be to omit the if and instead add .filter(Objects::nonNull) to the stream. Both are fine, I'll let you choose which one you find more understandable/readable.

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesMetricSamplesHandler.java
```diff
@@ -0,0 +1,188 @@
...
+        this.handledAttributes = Objects.requireNonNull(handledAttributes);
+        this.excludedAttributes = Objects.requireNonNull(excludedAttributes);
+        if (!handledAttributes.isEmpty() &&  !excludedAttributes.isEmpty()) {
+            throw new IllegalArgumentException("Either a set of handled attributes or a set of excluded attributes is required, setting both is not supported.");
+        }
+        MetricHeartbeatRegistry.getInstance().registerHandler(this);
+    }
+
+    @Override
+    public void initializeExecutionContext(ExecutionEngineContext executionEngineContext, ExecutionContext executionContext) {
```
cl-exense: Not sure if this can be omitted; if not, maybe define an empty default in the interface.

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesMetricSamplesHandler.java
```diff
@@ -0,0 +1,188 @@
...
+    private final TimeSeries timeSeries;
+
+    private final Set<String> handledAttributes;
+    private final Set<String> excludedAttributes;
+
+    public TimeSeriesMetricSamplesHandler(TimeSeries timeSeries, Set<String> handledAttributes, Set<String> excludedAttributes) {
+        this.timeSeries = timeSeries;
+        this.handledAttributes = Objects.requireNonNull(handledAttributes);
+        this.excludedAttributes = Objects.requireNonNull(excludedAttributes);
+        if (!handledAttributes.isEmpty() &&  !excludedAttributes.isEmpty()) {
```
cl-exense: I suppose that we also want to check that one of them is NOT empty -- in fact, that exactly one is empty and the other is not. That would be `if (!(a.isEmpty() ^ b.isEmpty())`. The error message would then become "exactly one of ... must be provided".

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesMetricSamplesHandler.java
```diff
@@ -0,0 +1,188 @@
...
+            handledAttributes.forEach(a -> {
+                if (measurement.containsKey(a)) {
+                    bucketAttributesMap.put(a, measurement.get(a));
+                }
+            });
+        }
+        return new BucketAttributes(bucketAttributesMap);
+    }
+
+    public void processThreadGroupAsMeasurement(Measurement measurement) {
```
cl-exense: That's a weird method name, I don't see anything related to thread groups. Also, this method does nothing if the argument is null. If this is indeed a public method, shouldn't it be the caller's responsibility to only invoke if not null? (There may be valid reasons for both,it just reads strange in isolation).

EDIT: Found the reference in this class; this seems to be the only usage, so I think this method should be `private void processThreadGroupMeasurement` and the null check is redundant and can be omitted.

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesMetricSamplesHandler.java
```diff
@@ -0,0 +1,188 @@
...
+        metrics.forEach(this::processMetric);
+    }
+
+    public void processMetric(AbstractMetricSample mm) {
+        MetricSample sample = mm.sample;
+        long begin = sample.getSampleTime();
+        BucketAttributes attributes = metricSampleToBucketAttributes(mm);
+        String instrumentType = sample.getType().toLowerCase();
+        attributes.put(METRIC_TYPE, mm.metricType != null ? mm.metricType : instrumentType);
+        attributes.put("instrumentType", instrumentType);
```
cl-exense: prefer using constants for well-known attribute names.

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesMetricSamplesHandler.java
```diff
@@ -0,0 +1,188 @@
...
+        measurement.remove("_id"); // because these measurements come with a generated id and can't be grouped into buckets.
+        if (measurement.getType().equals(THREAD_GROUP)) {
+            //Convert to the ThreadGroup Metric Sample
+            processThreadGroupAsMeasurement(measurement);
+        } else {
+            this.processMeasurement(measurement);
+        }
+    }
+
+    @Override
```
cl-exense: Same as above: empty default in interface?

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesService.java
```diff
@@ -87,30 +93,35 @@ public TimeSeriesAPIResponse getReportNodesTimeSeries(@NotNull FetchBucketsReque
...
-            return handler.getOrBuildTimeSeries(request);
+            return handler.fetchBucketsWithFallback(request);
         } catch (Exception e) {
             throw new ControllerServiceException(e.getMessage());
         }
     }
 
     private void enrichRequest(FetchBucketsRequest request) {
-        request.setOqlFilter(enrichOqlFilter(request.getOqlFilter(), request.isIncludeGlobalEntities()));
+        Boolean controllerMetric = Optional.ofNullable(request.getMetricType())
```
cl-exense: Very very minor: use boolean instead of Boolean, put .orElse on own line so the logic pipeline reads easily.

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesService.java
```diff
@@ -87,30 +93,35 @@ public TimeSeriesAPIResponse getReportNodesTimeSeries(@NotNull FetchBucketsReque
...
+            .map(metricType -> metricType.getCustomField(IS_CONTROLLER_METRIC, Boolean.class)).orElse(false);
+
+        request.setOqlFilter(enrichOqlFilter(request.getOqlFilter(), request.isIncludeGlobalEntities(), controllerMetric));
         if (request.getMaxNumberOfSeries() <= 0) {
             request.setMaxNumberOfSeries(maxNumberOfSeries);
         }
     }
 
-    private String enrichOqlFilter(String oqlFilter, boolean includeGlobalEntities) {
+    private String enrichOqlFilter(String oqlFilter, boolean includeGlobalEntities, Boolean controllerMetric) {
```
cl-exense: see above, use primitive boolean, there is no possibility for "undefined".

step-controller/step-controller-server/src/main/java/step/core/controller/StepControllerPlugin.java
```diff
@@ -34,8 +34,8 @@ public class StepControllerPlugin extends AbstractControllerPlugin {
 
     @Override
     public void init(GlobalContext context) throws Exception {
-        // Only used for validation here, blows up if there is something wrong with the configuration.
-        getControllerUrl(context.getConfiguration(), false, true);
+        context.setControllerUrl(getControllerUrl(context.getConfiguration(), false, true));
```
cl-exense, line 38: This will very probably conflict with the recent changes I did to move the validation to the earliest possible bootstrap stage which is now on master. To be tested, but I guess that it should be fine to just move this logic to the bootstrap method then. Also, the last argument (validate) can be set to false on the second call as it would be redundant.

step-controller/step-controller-server/src/main/java/step/core/controller/StepControllerPlugin.java
```diff
@@ -59,7 +59,7 @@ public void init(GlobalContext context) throws Exception {
      * @return the corresponding URL
      * @throws PluginCriticalException if the configuration is invalid.
      */
-    public static String getControllerUrl(Configuration conf, boolean forBackend, boolean validate) throws PluginCriticalException {
+    private String getControllerUrl(Configuration conf, boolean forBackend, boolean validate) throws PluginCriticalException {
```
cl-exense, line 67: In fact, if with is private now, the validate arg can be completely omitted and the check replaced by (!forBackend).

step-controller/step-controller-server/src/main/java/step/core/deployment/AbstractStepServices.java
```diff
@@ -90,6 +90,11 @@ protected ObjectFilter getRestrictedObjectFilter() {
         return objectHookRegistry.getObjectFilter(session);
     }
 
+    protected ObjectFilter getUnscopedObjectFilter() {
+        Session<User> session = new UnScopedSession(getSession());
```
cl-exense: Not sure if I mentioned it elsewhere - the class name should be Un**s**copedSession.

step-controller/step-controller-server/src/main/java/step/core/deployment/UnScopedSession.java
```diff
@@ -0,0 +1,44 @@
...
+ * When using this class as context for {@link step.core.objectenricher.ObjectHook} related operations,
+ * the scope of the operations is not restricted to tenant of entities.
+ * <p>
+ * This class has been introduced to perform {@code step.plugins.timeseries.TimeSeriesService} requests
+ * unrestricted to a specific project. The implementation should ensure that only entities not attached to a project can be accessed
+ * <p>
+ * In the future we might introduce a request parameter object to perform such operations without having to
+ * "misuse" the context
+ */
+public class UnScopedSession extends Session<User> {
```
cl-exense: UnScopedSession -> UnscopedSession

step-controller/step-controller-server/src/main/java/step/core/metrics/AbstractMetricSample.java
```diff
@@ -0,0 +1,54 @@
...
+ *   <li>{@link ControllerMetricSample} — carries no execution context; used for
+ *       controller-level metrics such as grid token usage.</li>
+ * </ul>
+ */
+@JsonTypeInfo(include = JsonTypeInfo.As.PROPERTY, use = JsonTypeInfo.Id.CLASS)
+@JsonSubTypes(value = {
+    @JsonSubTypes.Type(ControllerMetricSample.class),
+    @JsonSubTypes.Type(ExecutionMetricSample.class),
+})
+public abstract class AbstractMetricSample extends AbstractOrganizableObject {
```
cl-exense: Not happy with the name here, especially because this isn't an implementation of a metric sample, but an object containing a sample, for some specific purpose.

According to the javadoc and existing subclasses, I think this should be named something like
PersistableMetricSample, OrganizableMetricSample, TypedMetricSample, ExtendedMetricSample, SpecializedMetricSample, ConcreteMetricSample, StepMetricSample or something like that. Not 100% happy with either of them, but you get the idea.

step-controller/step-controller-server/src/main/java/step/core/metrics/AbstractMetricSample.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Optional time-series category override. When non-null, handlers that write a
+     * {@code metricType} attribute use this value instead of
+     * {@code sample.getType().toLowerCase()}.
+     */
+    public final String metricType;
+
+    protected AbstractMetricSample(MetricSample sample, String metricType) {
+        this.sample = sample;
```
cl-exense: Any non-null checks to perform here?

step-controller/step-controller-server/src/main/java/step/core/metrics/AbstractMetricSample.java
```diff
@@ -0,0 +1,54 @@
...
+        this.metricType = metricType;
+    }
+
+    /**
+     * Returns the flat label map that handlers should use when writing this metric to
+     * their respective storage.  The map includes both the developer-defined labels on
+     * the underlying {@link MetricSample} and any context-specific labels added by the
+     * concrete subclass (e.g. execution ID, agent URL).
+     */
+    public abstract Map<String, String> getEffectiveLabels();
```
cl-exense: Both implementations return a TreeMap - should this directly be enforced here?

step-controller/step-controller-server/src/main/java/step/core/metrics/ExecutionMetricSample.java
```diff
@@ -0,0 +1,94 @@
...
+     * <p>
+     * This is the full set of dimensions that handlers (e.g. time-series, Prometheus) should
+     * use when writing the metric to their respective storage.
+     * <p>
+     * User-defined labels take precedence only for keys not already set by the context.
+     * Context labels are set last so they cannot be overridden by user labels.
+     */
+    @JsonIgnore
+    @Override
+    public Map<String, String> getEffectiveLabels() {
```
cl-exense: Proposed small change:

```
    public Map<String, String> getEffectiveLabels() {
        TreeMap<String, String> labels = new TreeMap<>(sample.getLabels());
        putIfNotEmpty(labels, MetricsExecutionPlugin.ORIGIN, origin);
        if (attributes != null) {
            labels.putAll(attributes);
        }
        // Context labels are authoritative — set last so they cannot be overridden
        putIfNotEmpty(labels, MetricsExecutionPlugin.ATTRIBUTE_EXECUTION_ID, eId);
        putIfNotEmpty(labels, MetricsExecutionPlugin.PLAN_ID, planId);
        putIfNotEmpty(labels, MetricsExecutionPlugin.PLAN, plan);
        putIfNotEmpty(labels, MetricsExecutionPlugin.AGENT_URL, agentUrl);
        putIfNotEmpty(labels, MetricsExecutionPlugin.TASK_ID, taskId);
        putIfNotEmpty(labels, MetricsExecutionPlugin.SCHEDULE, schedule);
        putIfNotEmpty(labels, MetricsExecutionPlugin.EXECUTION_DESCRIPTION, execution);
        return labels;
    }

    private void putIfNotEmpty(Map<String, String> map, String key, String value) {
        if (value != null && !value.isEmpty()) {
            map.put(key, value);
        }
    }
```

This reads slightly easier. It minimally changes the semantics as it does the same non-empty check on all values, but I think empty values don't make sense for origin etc. either (correct me if I'm wrong).

step-controller/step-controller-server/src/main/java/step/core/metrics/MetricHeartbeatRegistry.java
```diff
@@ -0,0 +1,155 @@
...
+
+    public static MetricHeartbeatRegistry getInstance() {
+        return INSTANCE;
+    }
+
+    private final ConcurrentHashMap<String, LastSampleEntry> lastSamples = new ConcurrentHashMap<>();
+    private final List<MetricSamplesHandler> handlers = new CopyOnWriteArrayList<>();
+    private final ScheduledExecutorService scheduler =
+            Executors.newScheduledThreadPool(1, new BasicThreadFactory.Builder()
+                    .namingPattern("metric-heartbeat-%d").build());
```
cl-exense: add a .daemon() call to the builder so it returns daeomn threads.

step-controller/step-controller-server/src/main/java/step/core/metrics/MetricHeartbeatRegistry.java
```diff
@@ -0,0 +1,155 @@
...
+    volatile long intervalMs;
+
+    private static final class LastSampleEntry {
```
cl-exense, line 45: Let's start using the new features :-)

```
private record LastSampleEntry(ExecutionMetricSample sample, long lastUpdatedMs) {
}
```

step-controller/step-controller-server/src/main/java/step/core/metrics/MetricHeartbeatRegistry.java
```diff
@@ -0,0 +1,155 @@
...
+                    }
+                }
+            }
+        } catch (Exception e) {
+            logger.error("Exception in metric heartbeat tick", e);
+        }
+    }
+
+    private static String buildKey(ExecutionMetricSample mm) {
+        String execId = mm.eId != null ? mm.eId : "";
```
cl-exense: more readable (this and following lines):
```
String execId = Objects.requireNonNullElse(mm.eId, "");
```

step-controller/step-controller-server/src/main/java/step/core/metrics/MetricHeartbeatRegistry.java
```diff
@@ -0,0 +1,155 @@
...
+        String agentUrl = mm.agentUrl != null ? mm.agentUrl : "";
+        String origin = mm.origin != null ? mm.origin : "";
+        Map<String, String> labels = mm.sample.getLabels();
+        String labelsKey = labels != null ? new TreeMap<>(labels).toString() : "{}";
+        return execId + "|" + agentUrl + "|" + origin + "|" + mm.sample.getName() + "|" + labelsKey;
```
cl-exense: More readable:
```
return String.join("|", execId, agentUrl, origin, mm.sample.getName(), labelsKey);
```

step-controller/step-controller-server/src/main/java/step/core/metrics/MetricHeartbeatRegistry.java
```diff
@@ -0,0 +1,155 @@
...
+     *   <li><b>GAUGE</b>: {@code count=1, sum/min/max/last = lastObservedValue} — repeats
+     *       the last known value as a single synthetic observation.</li>
+     * </ul>
+     */
+    private static ExecutionMetricSample buildHeartbeat(ExecutionMetricSample original, long now) {
+        MetricSample orig = original.sample;
+        MetricSample heartbeatSample;
+        if (orig.getType() == InstrumentType.COUNTER) {
+            long runningTotal = orig.getSum();
+            heartbeatSample = new MetricSample(now, orig.getName(), orig.getLabels(), InstrumentType.COUNTER,
```
cl-exense, line 143: Ah, this looks familiar :-) -- there are some potential changes in the API, so I think they would also affect the logic here.

step-controller/step-controller-server/src/main/java/step/core/metrics/MetricSampler.java
```diff
@@ -0,0 +1,19 @@
+package step.core.metrics;
+
+import java.util.List;
+
+public abstract class MetricSampler {
+    public String name;
+    public String description;
+
+    public MetricSampler(String name, String description) {
+        this.name = name;
```
cl-exense: Non-null checks? I know I'm being anal...

step-controller/step-controller-server/src/main/java/step/core/metrics/MetricSamplerRegistry.java
```diff
@@ -0,0 +1,66 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(MetricSamplerRegistry.class);
+
+    private static final MetricSamplerRegistry INSTANCE = new MetricSamplerRegistry();
+
+    public static MetricSamplerRegistry getInstance() {
+        return INSTANCE;
+    }
+
+    private final ScheduledExecutorService scheduler =
+        Executors.newScheduledThreadPool(1, BasicThreadFactory.builder().namingPattern("metric-sampler-%d").build());
```
cl-exense: use daemon threads.

step-controller/step-controller-server/src/main/java/step/core/metrics/MetricSamplerRegistry.java
```diff
@@ -0,0 +1,66 @@
...
+    public void registerSampler(String name, MetricSampler sampler) {
+        samplers.put(name, sampler);
+    }
+
+    public void registerHandler(MetricSamplesHandler handler) {
+        handlers.add(handler);
+    }
+
+    public void start(int interval) {
+        Runnable collect = () -> {
```
cl-exense: Move this to its own method as you did in MetricHeartbeatRegistry, that's more understandable.

step-controller/step-controller-server/src/main/java/step/core/metrics/MetricsControllerPlugin.java
```diff
@@ -0,0 +1,212 @@
...
+    protected void initMetricSamplingAndHeartbeat(GlobalContext context) {
+        metricSamplerRegistry = MetricSamplerRegistry.getInstance();
+
+        //Start the metric sampling scheduler
+        int interval = context.getConfiguration().getPropertyAsInteger("plugins.measurements.gaugecollector.interval", 15);
+        metricSamplerRegistry.start(interval);
+
+        //Start the metric heartbeat scheduler
+        int heartbeatIntervalSec = context.getConfiguration().getPropertyAsInteger("plugins.metrics.metricheartbeat.interval", 15);
+        MetricHeartbeatRegistry.getInstance().start(heartbeatIntervalSec * 1000L);
```
cl-exense, line 98: For consistency, both registries should either accept milliseconds, or seconds, as intervals. Oh, it's also a good idea to name the arguments accordingly (i.e. intervalMs, intervalSeconds).

step-plans/step-plans-base-artefacts/src/main/java/step/livereporting/LiveReportingContext.java
```diff
@@ -32,6 +33,7 @@ public class LiveReportingContext {
      */
     public final String id = UUID.randomUUID().toString();
     private final List<LiveReportingContextListener> listeners = new ArrayList<>();
```
cl-exense: We should probably rename the classes to LiveReportingContextMeasureListener and LiveReportingContextMetricListener (and rename field and method names accordingly).

step-controller/step-controller-server/src/main/java/step/core/metrics/MetricsExecutionPlugin.java
```diff
@@ -0,0 +1,488 @@
...
+
+        processMetrics(List.of(stepSample));
+        // Thread group metrics span the whole execution lifetime (unlike keyword metrics),
+        // so they are eligible for heartbeat re-emission when no new value arrives.
+        // Synchronized on counter to serialize heartbeat updates: concurrent threads may
+        // complete their atomic increment/decrement in order but reach update() in a
+        // different order, leaving the heartbeat with a stale count (e.g. count=1
+        // overwriting count=3). Re-reading counter.get() inside the lock ensures the
+        // heartbeat always reflects the true current count regardless of which thread wins.
+        synchronized (counter) {
```
cl-exense, line 173: I didn't entirely get the complete rationale, but yeah, better safe than sorry. However, in this case, it seems prudent to also put the updates into a synchronized block (line 166)) ? Otherwise, the values here might still be stale because they can still change while the lock is held. In the worst case, we're wasting a few milliseconds I guess.

step-controller/step-controller-server/src/test/java/step/core/metrics/MetricsExecutionPluginTest.java
```diff
@@ -49,18 +48,18 @@
...
     private static Map<String, AtomicInteger> assertMeasurementsCount = new HashMap<>();
     private static Map<String, AtomicLong> assertMeasurementsValue = new HashMap<>();
 
     @Before
     public void setUp() throws Exception {
-        MeasurementControllerPlugin mc = new MeasurementControllerPlugin();
-        mc.initGaugeCollectorRegistry(GlobalContextBuilder.createGlobalContext());
-        MeasurementPlugin.registerMeasurementHandlers(new TestMeasurementHandler());
-        engine = ExecutionEngine.builder().withPlugin(new MeasurementPlugin(GaugeCollectorRegistry.getInstance()))
+        step.core.metrics.MetricsControllerPlugin mc = new step.core.metrics.MetricsControllerPlugin();
```
cl-exense: Unneeded fully-qualified class names used in this class; use imports instead.

step-controller/step-controller-server/src/test/java/step/core/GlobalContextBuilder.java
```diff
@@ -129,6 +129,9 @@ public static GlobalContext createGlobalContext() throws CircularDependencyExcep
                 resourceRevisionAccessor, ResourceRevision.class));
 
         context.setArtefactHandlerRegistry(new ArtefactHandlerRegistry());
+
```
cl-exense, line 132: Why is this required? The Controller.java also creates one using the "normal" collection factory, supposedly overriding the one set here? Can this cause any issues?

step-livereporting/step-livereporting-client-impl/src/main/java/step/livereporting/client/RestUploadingLiveMetricDestination.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    public RestUploadingLiveMetricDestination(String endpointUrl, int batchSize, long flushIntervalMs) {
+        this.endpointUrl = endpointUrl;
+        this.client = createClient();
+        this.batchProcessor = new BatchProcessor<>(batchSize, flushIntervalMs, this::sendMetrics, "livereporting-metrics-rest");
+        this.metricSamplesBuilder = new MetricSamplesBuilder(batchProcessor::add);
+    }
+
+    private Client createClient() {
```
cl-exense, line 72: Optional: Should we create a base class, say RestUploadingLiveReportingDestination and refactor the common code like createClient() , and potentially even the shared batchprocessing logic and sending logic, into it?

step-plans/step-plans-core/src/main/java/step/core/execution/AbstractExecutionEngineContext.java
```diff
@@ -34,6 +34,9 @@
 
 public abstract class AbstractExecutionEngineContext extends AbstractStepContext {
 
+    public static final String CONTROLLER_URL_KEY = "CONTROLLER_URL_KEY";
```
cl-exense: These should probably be private.

cl-exense (changes requested):
OK, second part of review, I tried to understand most of the newly added code but for instance only skimmed over the unit tests; few comments are directly related to logic, most focus on making the code safer or easier to read, and some are more at an architectural level.

step-controller-plugins/step-controller-plugins-metrics-raw/src/main/java/step/plugins/metrics/raw/MeasurementAccessor.java
david-stephan: nope, this is still the MeasurementAccessor to do CRUD operations on Measurements.

step-controller-plugins/step-controller-plugins-metrics-raw/src/main/java/step/plugins/metrics/raw/MetricSampleAccessor.java
david-stephan: You're right, I initially thought to make it generic for metric samples but we anyway do not (want to) persist controller level metrics as RAW data and the implementation is not generic, so refactoring the name

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesControllerPlugin.java
david-stephan: I think it was even worse before this refactoring, but you're right I now completely removed it from the logs.

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesMetricSamplesHandler.java
david-stephan: It is allowed to have both empty, I'll add a java doc to the constructor

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesMetricSamplesHandler.java
david-stephan: did further refactoring, but I should really stop at some point....

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesMetricSamplesHandler.java
david-stephan: indeed

step-controller/step-controller-server/src/main/java/step/core/metrics/AbstractMetricSample.java
david-stephan: make sense on sample. type can be null

step-controller/step-controller-server/src/main/java/step/core/metrics/ExecutionMetricSample.java
david-stephan: technically it would be possible to have an empty string as plan name and search for it, but applied the proposed change as this is rather a edge case. Most critical point it so keep a constant ordered set  of labels for some integration, but this is now directly handled in the respective integration implementation.

step-controller/step-controller-server/src/main/java/step/core/metrics/MetricSampler.java
david-stephan: to be honest I don't think those fields are ever used but why not.

step-controller/step-controller-server/src/main/java/step/core/metrics/MetricsControllerPlugin.java
david-stephan: aligned both impl.

step-controller/step-controller-server/src/main/java/step/core/metrics/MetricsExecutionPlugin.java
david-stephan: yes, didn't try to rethink it too much, it's definitely safer to start the synchronize block line 166

step-controller/step-controller-server/src/test/java/step/core/GlobalContextBuilder.java
david-stephan: This class is strictly used by Junit tests (not starting a controller), Javadoc should be probably added to this class, but this is nothing new.

step-plans/step-plans-base-artefacts/src/main/java/step/livereporting/LiveReportingContext.java
david-stephan: I wasn't too sure where the legacy class and method names were used, I refactored now, hopefully that doesn't break anything

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesMetricSamplesHandler.java
cl-exense: Can still be removed :-D

cl-exense (dismissed):
Approving with one final comment (unnecessary empty override of default method)

PR #635, opened 2026-04-17
SED-4655 Support .mjs/.cjs keyword modules and fix keyword detection for CJS modules with indirect module.exports assignment 

jeromecomte:


david-stephan (dismissed):
Approving, but gemini comment makes sense

david-stephan (approved):

PR #636, opened 2026-04-17
SED-4654 Ensure proper propagation of uncaught exceptions, unhandled rejections, and module import failures

jeromecomte:
…rejections, and module import failures

### DIFF
--- a/step-node/step-node-agent/api/controllers/agent-fork.js
+++ b/step-node/step-node-agent/api/controllers/agent-fork.js
@@ -77,7 +77,7 @@ process.on('message', async ({ type, projectPath, functionName, input, propertie
       }
     } catch (e) {
       console.log("[Agent fork] Unexpected error occurred while executing keyword", e)
-      outputBuilder.fail("An unexpected error occurred while executing keyword: " + e.message, e)
+      outputBuilder.fail("An unexpected error occurred while executing keyword: " + (e?.message || String(e)), e)
     } finally {
       // Flush the event loop so unhandledRejection / uncaughtException from the keyword
       // (e.g. fire-and-forget promises, nextTick throws) land before we send the result.
@@ -120,7 +120,19 @@ process.on('message', async ({ type, projectPath, functionName, input, propertie
     console.log("[Agent fork] Exiting...")
     if (closeErrors.length > 0) {
       // Use the send callback to ensure the message is flushed before we exit.
-      process.send({ type: 'CLOSE_RESULT', errors: closeErrors }, () => process.exit(0));
+      // The timeout is a safety net: if the IPC channel closes before the callback
+      // fires (e.g. the parent process dies), the fork exits instead of hanging.
+      const exitNow = () => process.exit(0);
+      const fallback = setTimeout(exitNow, 5000);
+      try {
+        process.send({ type: 'CLOSE_RESULT', errors: closeErrors }, () => {
+          clearTimeout(fallback);
+          exitNow();
+        });
+      } catch {
+        clearTimeout(fallback);
+        exitNow();
+      }
     } else {
       process.exit(0);
     }
@@ -143,7 +155,7 @@ process.on('message', async ({ type, projectPath, functionName, input, propertie
           let module = await import(kwModule);
           kwModules.push(module);
         } catch (e) {
-          throw new Error("Error while importing keyword module " + kwFile + ": " + e.message, { cause: e })
+          throw new Error("Error while importing keyword module " + kwFile + ": " + (e?.message || String(e)), { cause: e })
         }
       }
     }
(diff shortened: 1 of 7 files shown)


step-node/step-node-agent/api/controllers/agent-fork.js
```diff
@@ -68,15 +75,55 @@ process.on('message', async ({ type, projectPath, functionName, input, propertie
...
+      // Flush the event loop so unhandledRejection / uncaughtException from the keyword
+      // (e.g. fire-and-forget promises, nextTick throws) land before we send the result.
+      await new Promise(resolve => setImmediate(resolve));
+      // Surface inter-keyword errors first, labelled clearly as coming from a previous keyword.
+      if (prevUnhandledRejection) {
+        const sep = outputBuilder.hasError() ? '\n' : '';
+        outputBuilder.appendError(`${sep}Unhandled promise rejection from a previous keyword: ${prevUnhandledRejection.message || prevUnhandledRejection}`);
+      }
+      if (prevUncaughtException) {
+        const sep = outputBuilder.hasError() ? '\n' : '';
```
david-stephan, line 91: Can it be that the output has no error, but we had a prevUnhandledRejection? In that case we should also add a line break separator. Same comment for the following usages?

david-stephan (changes requested):

step-node/step-node-agent/api/controllers/agent-fork.js
jeromecomte: After calling appendError, the method hasError returns true. So this case is also handled

step-node/step-node-agent/api/controllers/agent-fork.js
```diff
@@ -22,9 +22,16 @@ const Session = require("./session");
 const fs = require("fs");
 const path = require('path')
 const session = new Session();
+let pendingUnhandledRejection = null;
+let pendingUncaughtException = null;
```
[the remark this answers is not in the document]
jeromecomte: Agree but it's an edge case. For simplicity we'll keep as it is. The most important point is to notify the user when at least one uncaught error occur. Details can also be found in the logs

step-node/step-node-agent/api/controllers/agent-fork.js
```diff
@@ -68,15 +75,55 @@ process.on('message', async ({ type, projectPath, functionName, input, propertie
...
+      console.error('[Agent fork] Uncaught exception occurred after last keyword:', pendingUncaughtException);
+      closeErrors.push({ message: 'An uncaught exception occurred after the last keyword call: ' + (pendingUncaughtException.message || String(pendingUncaughtException)) });
+    }
     console.log("[Agent fork] Releasing session...")
     await session.asyncDispose();
     console.log("[Agent fork] Exiting...")
-    process.exit(0)
+    if (closeErrors.length > 0) {
+      // Use the send callback to ensure the message is flushed before we exit.
+      process.send({ type: 'CLOSE_RESULT', errors: closeErrors }, () => process.exit(0));
```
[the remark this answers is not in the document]
jeromecomte: Good point. Fallback implemented

step-node/step-node-agent/api/controllers/agent.js
```diff
@@ -441,10 +446,20 @@ class ForkedAgent {
...
         resolve();
       } else {
+        // Listen for close errors reported by the fork before it exits.
+        const messageListener = (msg) => {
+          if (msg && msg.type === 'CLOSE_RESULT') {
+            closeErrors = msg.errors;
+            this.forkProcess.removeListener('message', messageListener);
+          }
+        };
+        this.forkProcess.removeAllListeners('message');
```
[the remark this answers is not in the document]
jeromecomte: agree but at this stage it doesn't matter

david-stephan (approved):

step-node/step-node-agent/api/controllers/agent.js
```diff
@@ -441,10 +446,20 @@ class ForkedAgent {
...
   async close() {
-    const exitPromise = new Promise(resolve => {
+    let closeErrors = null;
+    const closePromise = new Promise(resolve => {
       if (this.forkProcess.exitCode !== null) {
         resolve();
       } else {
+        // Listen for close errors reported by the fork before it exits.
+        const messageListener = (msg) => {
+          if (msg && msg.type === 'CLOSE_RESULT') {
```
neogucky, line 456:
```suggestion
          if (msg?.type === 'CLOSE_RESULT') {
```

msg && .. is the very old JS way, in ts (or modern JS) we would use the ?. operator that checks for undefined before resolving the property. I also see this used elsewhere in the file.

step-node/step-node-agent/api/controllers/agent.js
```diff
@@ -453,8 +468,11 @@ class ForkedAgent {
     } catch {
       this.forkProcess.kill();
     }
-    await exitPromise;
+    await closePromise;
     fs.rmSync(this.agentForkerLibPath, {recursive: true, force: true});
+    if (closeErrors && closeErrors.length > 0) {
```
neogucky, line 473:
```suggestion
    if (closeErrors?.length > 0) {
```

neogucky (approved):

PR #642, opened 2026-05-06
SED-4673 Graceful handling of inexistent fragment references in AP

cl-exense:
Please see associated discussion/description in related ticket.

Also note that this is a PR against SED-4678 because it requires the changes from there. If I made it against master as base branch, it would be cluttered with all the changes from that branch too.


step-core/src/main/java/step/automation/packages/ResourcePathMatchingResolver.java
```diff
@@ -37,11 +38,16 @@ public ResourcePathMatchingResolver(ClassLoader classLoader) {
     public List<URL> getResourcesByPattern(String resourcePathPattern) {
         List<URL> res = new ArrayList<>();
         if (!containsWildcard(resourcePathPattern)) {
-            res.add(classLoader.getResource(resourcePathPattern));
+            URL url = classLoader.getResource(resourcePathPattern);
```
david-stephan, line 41: @cl-exense This methods is used by the AP readers not just by the new AP collections. Since the method is called getResourcesByPattern I assume it is fine to return an empty list if URL is null also for patterns without wildcard, but we should verify the usages.

cl-exense: The problem that this PR tries to solve is a NullPointerException when the resourcePathPattern is wrong, i.e. doesn't exist (independent of whether it contains a wildcard or not). This here is the simple case without a wildcard, for instance "nonexistentFile.yml" . As said, the current code will simply throw an NPE, while the proposed change would ignore it.

(Same for the change just below, except that one involves wildcards)

I assume that the other uses of this class until now are also simply expecting the happy path, and would fail when an NPE is thrown here. Conversely, ignoring "wrong declarations" would at least not break them.

david-stephan: Other use cases are not expecting the happy path, or at least one cannot just assume it. Deploying an AP specifying a fragment or a plain text plan file  (also used in this context) with a file that doesn't exists currently will fail and this is the expected behavior; that should not be changed.

david-stephan (changes requested):
see details in comments

david-stephan (approved):

PR #643, opened 2026-05-07
SED-4681 Process automation-package.yaml once, not twice

cl-exense:


step-automation-packages/step-automation-packages-collections/src/test/resources/testdata/ap1/automation-package.yml
```diff
@@ -12,7 +12,10 @@ alertingRules:
           predicate:
             BindingValueEqualsPredicate:
               value: "myValue"
-plans: []
+plans: [ ]
+parameters:
+  - key: "paramInMainAP"
+  - value: "once"
```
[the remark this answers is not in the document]
cl-exense: Already fixed.

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
```diff
@@ -92,7 +88,9 @@ public AutomationPackageYamlFragmentManager(AutomationPackageDescriptorYaml desc
         this.stagingContext = stagingContext;
         initializeMaps(descriptorYaml);
 
-        pathToYamlFragment.values().forEach(this::initializeMaps);
+        pathToYamlFragment.values().stream()
+            .filter(f -> f != descriptorYaml)
+            .forEach(this::initializeMaps);
```
cmisev, line 93: I see. That's quite an obvious mistake of mine. However easier might be to just remove initializeMaps(descriptorYaml)?

Raises the further question if we need descriptorYaml as a separate field at all, I initially had it because I was more ambitious at the beginning of how quickly I implement support for adding and removing fragment references. Let's keep it for now.

cmisev (changes requested):
See line comment. I unless there is something I missed, instead of using the filter you could simply remove the call above to initializeMaps(descriptorYaml)

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
cl-exense: I had thought of simply removing the first "standalone" initializeMaps call too, and it seems to essentially have the same effect. I still prefer to actually keep it this way, just to ensure that the main descriptor always gets processed first in case if has/will have any effect that might be required before processing the rest. We can always change it later on, it's a trivial change.

cmisev: ok, then I approve

cmisev (approved):

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
cmisev: (can't think of any side effect, but let's keep on the cautous side)

PR #644, opened 2026-05-13
SED-4617 implement-detailed-inline-node-details-mode-for-step-list

david-stephan:


step-controller/step-controller-server/src/main/java/step/core/execution/table/ReportNodeTableEnricher.java
```diff
@@ -0,0 +1,45 @@
...
+            && ((ReportNodesTableParameters) tableParameters).isEnrichWithContributingErrors()
+            && (!reportNode.isLeafReportNode())
+            && ((ReportNodeStatus.FAILED.equals(reportNode.getStatus())) || (ReportNodeStatus.TECHNICAL_ERROR.equals(reportNode.getStatus())));
+    }
+
+    private List<ReportNode> collectContributingErrors(ReportNode root) {
+        try (Stream<ReportNode> reportNodesWithContributingErrors = reportNodeAccessor.getReportNodesWithContributingErrorsByAncestor(root.getId().toHexString(), 0, 10)) {
+            return reportNodesWithContributingErrors.collect(Collectors.toList());
+        }
+    }
```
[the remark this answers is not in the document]
david-stephan: The n+1 problematic is known. With the current access layer compatible with different DB backend including NO SQL it is not possible to perform join queries and this remains the only implementation. This why this it only done for specific nodes and only when explicitly requested with the table parameter flag enrichWithContributingErrors. A new field enrichWithContributingErrorsLimit with a default value of 10 has been added to ReportNodesTableParameters allowing the configuration of the limit per request.

step-controller/step-controller-server/src/main/java/step/core/execution/ReportNodesFilter.java
```diff
@@ -39,10 +41,13 @@ public List<Filter> buildAdditionalQuery(ReportNodesTableParameters parameters)
                 fragments.add(Filters.equals("executionID", eid));
```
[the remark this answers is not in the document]
david-stephan: done

step-controller/step-controller-server/src/main/java/step/migration/tasks/V30_0_ReportNodeAncestorIdsMigrationTask.java
```diff
@@ -0,0 +1,82 @@
...
+        try (Stream<Document> documents = reports.findLazy(Filters.exists(PATH_FIELD_NAME), null, null, null, 0)) {
+            documents.forEach(document -> {
+                Object pathObject = document.remove(PATH_FIELD_NAME);
+                if (pathObject instanceof String path) {
+                    List<String> ancestorIds = new ArrayList<>();
+                    for (int i = 0; i + OBJECT_ID_HEX_LENGTH <= path.length(); i += OBJECT_ID_HEX_LENGTH) {
+                        ancestorIds.add(path.substring(i, i + OBJECT_ID_HEX_LENGTH));
+                    }
+                    if (!ancestorIds.isEmpty()) {
+                        document.put("ancestorIds", ancestorIds);
```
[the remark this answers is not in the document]
david-stephan: Used a new constant declared in this migration task directly, the migration tasks should not depend on the current model but stick to the model corresponding to the version for which it applies to.

step-controller/step-controller-server/src/main/java/step/migration/tasks/V30_0_ReportNodeAncestorIdsMigrationTask.java
```diff
@@ -0,0 +1,82 @@
...
+                    }
+                }
+                reports.save(document);
+                counter.incrementAndGet();
+            });
```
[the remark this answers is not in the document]
david-stephan: Switching to a batch implementation, but this will only improve the processing with PSQL backends, MongoDB implementation still fall back to individual upserts

step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/ReportNodeAccessorImpl.java
```diff
@@ -36,20 +38,32 @@
...
+        createOrUpdateIndex(EXECUTION_TIME_FIELD_NAME);
+        createOrUpdateCompoundIndex(EXECUTION_ID_FIELD_NAME, STATUS_FIELD_NAME, EXECUTION_TIME_FIELD_NAME);
+        createOrUpdateCompoundIndex(EXECUTION_ID_FIELD_NAME, EXECUTION_TIME_FIELD_NAME);
+        createOrUpdateCompoundIndex(EXECUTION_ID_FIELD_NAME, CLASS_FIELD_NAME);
+        createOrUpdateCompoundIndex(EXECUTION_ID_FIELD_NAME, PARENT_ID_FIELD_NAME);
+        createOrUpdateCompoundIndex(EXECUTION_ID_FIELD_NAME, ARTEFACT_HASH_FIELD_NAME);
+
+        IndexField indexFieldExecutionId = new IndexField(EXECUTION_ID_FIELD_NAME, Order.ASC, String.class);
+        IndexField indexFieldAncestorIds = new IndexField(ANCESTOR_IDS_FIELD_NAME, Order.ASC, List.class);
+        createOrUpdateCompoundIndex(new LinkedHashSet<>(Set.of(indexFieldExecutionId, indexFieldAncestorIds)));
```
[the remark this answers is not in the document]
david-stephan: fixed

step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/ReportNodeAccessorImpl.java
```diff
@@ -72,35 +86,35 @@ private void appendParentNodeToPath(LinkedList<ReportNode> path, ReportNode node
...
     @Override
     public Iterator<ReportNode> getChildren(ObjectId parentID, int skip, int limit) {
-        return collectionDriver.find(Filters.equals("parentID", parentID), new SearchOrder("executionTime", 1), skip, limit, 0).iterator();
+        return collectionDriver.find(Filters.equals(PARENT_ID_FIELD_NAME, parentID), new SearchOrder(EXECUTION_TIME_FIELD_NAME, 1), skip, limit, 0).iterator();
     }
 
     @Override
     public Iterator<ReportNode> getChildrenByParentSource(ObjectId parentID, ParentSource parentSource) {
-        return collectionDriver.find(Filters.and(List.of(Filters.equals("parentID", parentID), Filters.equals("parentSource", parentSource.name()))), new SearchOrder("executionTime", 1), null, null, 0).iterator();
+        return collectionDriver.find(Filters.and(List.of(Filters.equals(PARENT_ID_FIELD_NAME, parentID), Filters.equals("parentSource", parentSource.name()))), new SearchOrder(EXECUTION_TIME_FIELD_NAME, 1), null, null, 0).iterator();
```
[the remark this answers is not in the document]
david-stephan: done

step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/ReportNodeAccessorImpl.java
```diff
@@ -213,21 +227,28 @@ public Iterator<ReportNode> getChildrenByParentSource(String parentID, ParentSou
...
+        return collectionDriver.findLazy(Filters.and(contributingErrorFilters), null, skip, limit, 0);
     }
 
+    @Override
+    public Stream<ReportNode> getReportNodesWithContributingErrorsByAncestor(String ancestorId, Integer skip, Integer limit) {
+        Objects.requireNonNull(ancestorId);
+        List<Filter> contributingErrorFilters = new ArrayList<>(List.of(Filters.includes(ANCESTOR_IDS_FIELD_NAME, ancestorId),
+            Filters.equals(CONTRIBUTING_ERROR_FIELD_NAME, true)));
+        return collectionDriver.findLazy(Filters.and(contributingErrorFilters), null, skip, limit, 0);
+    }
```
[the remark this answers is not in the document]
david-stephan: This was indeed an issue and now executionID is enforced, indexes were as well reworked and optimized.

step-controller/step-controller-server/src/main/java/step/core/deployment/ControllerServices.java
```diff
@@ -166,6 +171,22 @@ public List<ReportNode> getReportNodeChildren(@PathParam("id") String reportNode
         return result;
     }
 
+    @GET
+    @Path("/reportnode/{id}/reportnodes-with-errors")
```
jeromecomte: In this context I would probably rename it to descendants-with-errors

step-controller/step-controller-server/src/main/java/step/core/execution/table/ReportNodeTableEnricher.java
```diff
@@ -0,0 +1,48 @@
...
+                return new EnrichedReportNode<>(reportNode, contributingErrors);
+            }
+        }
+        return reportNode;
+    }
+
+    private static boolean shouldEnrich(ReportNode reportNode, ReportNodesTableParameters tableParameters) {
+        return tableParameters.isEnrichWithContributingErrors()
+            && (!reportNode.isLeafReportNode())
+            && ((ReportNodeStatus.FAILED.equals(reportNode.getStatus())) || (ReportNodeStatus.TECHNICAL_ERROR.equals(reportNode.getStatus())));
```
jeromecomte, line 40: Just realized something while reading this: in the legacy view, if I'm not wrong, we were also displaying assertions of keyword calls for passed calls. In this case we wouldn't show them anymore.

Regardless of the legacy view, I'm asking me if showing the executed assertions in passed calls couldn't be required in some cases for transparency / auditing reasons

step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/handlers/EchoHandler.java
```diff
@@ -42,6 +42,8 @@ protected void execute_(EchoReportNode node, Echo testArtefact) {
 
     @Override
     public EchoReportNode createReportNode_(ReportNode parentNode, Echo testArtefact) {
-        return new EchoReportNode();
+        EchoReportNode reportNode = new EchoReportNode();
+        reportNode.setLeafReportNode(true);
```
jeromecomte: As this is static, it would be better to store this in the metadata of the Artefact annotation and access it via reportNode.resolvedArtefact

step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/aggregated/ReportNodeTimeSeries.java
```diff
@@ -72,8 +72,10 @@ public void ingestReportNode(ReportNode reportNode, Map<String, Object> customAt
         nodeBucket.put(ARTEFACT_HASH, reportNode.getArtefactHash());
         nodeBucket.put(STATUS, status.toString());
         if (reportNode.getError() != null) {
-            nodeBucket.put(ERROR_CODE, reportNode.getError().getCode());
-            nodeBucket.put(ERROR_MESSAGE, reportNode.getError().getMsg());
+            nodeBucket.put(ERROR_CODE, reportNode.getError().getCode() != null ? reportNode.getError().getCode() : 0);
```
jeromecomte, line 75: If 0 should be the default error code, shouldn't we define this default in the Error class?

david-stephan: @jeromecomte it definitely should have been, but is currently hardcoded all ove Step code base. I can do it but it will require a branch on the Step KW API then.

step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/handlers/EchoHandler.java
david-stephan: I did it here because the usage is more reporting related for now, I also thought that's what we discussed. We decided to let the handler set the value because the current model is shaky as any artefact can technically have children even if they are completely ignored by the execution of the handler.
I'm fine moving it to AbstractArtefact and its implemeting classes but in that case I would just have a default value in the abstract class and overrides with a JsonIgnore has it doesn't have to be persisted to DB. 
What do you think ? and what did you mean  with Artefact "annotation".

step-controller/step-controller-server/src/main/java/step/core/deployment/ControllerServices.java
david-stephan: renamed

step-controller/step-controller-server/src/main/java/step/core/execution/table/ReportNodeTableEnricher.java
david-stephan: I wasn't aware of this feature, but indeed if you expand the callKeyword report in the legacy step table the children are retrieved and the passed assertion are displayed too; this was OK in term of performance as only done on demands when expending one report node. 
Whether this feature is a requirement, I cannot tell, but I would definitely not use the table enricher for that as it would perform additional queries for each rows (or at least it should be optional too). Should that be a requirement, I would suggest to create a dedicated request for change and see how to best make this information available.

jeromecomte: Agree. Let's discuss this as a follow-up task

step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/handlers/EchoHandler.java
jeromecomte: I remember the discussion but though about it again.
If we later want to define some artifacts as leaf-only, the best approach would be to store this information in the step.core.artefacts.Artefact annotation (not the AbstractArtefact itself). This way we could also use the information in the plan editor / validation to enforce it.

step-plans/step-plans-core/src/main/java/step/core/artefacts/reports/aggregated/ReportNodeTimeSeries.java
jeromecomte: fine for me too if we want to do it later

step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/handlers/EchoHandler.java
david-stephan: I completely forgot that we have this annotation, I'll implement as  suggested

jeromecomte (dismissed):

PR #646, opened 2026-05-15
SED-4693-timeseries-ingest-begin-and-value-as-attributes

david-stephan:


step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesControllerPlugin.java
```diff
@@ -84,8 +89,11 @@ public void serverStart(GlobalContext context) {
         Set<String> excludedAttributes = Arrays.stream(configuration.getProperty(TIME_SERIES_EXCLUDED_ATTRIBUTES_PROPERTY, TIME_SERIES_EXCLUDED_ATTRIBUTES_DEFAULT).split(","))
             .filter(s -> !s.isEmpty())
             .collect(Collectors.toSet());
-
-        if (!includedAttributes.isEmpty() && !excludedAttributes.isEmpty()) {
+        if (includedAttributes.isEmpty()) {
+            //add mandatory exclude attributes
+            excludedAttributes.add(BEGIN);
+            excludedAttributes.add(VALUE);
+        } else if (!excludedAttributes.isEmpty()) {
```
jeromecomte: Shouldn't we always exclude BEGIN and VALUE and not just when includedAttributes are defined?

david-stephan: @jeromecomte I did not get your comment at first,  but we should indeed remove them from the included list to be "safe", I also added a bit more comment on the 2 operation modes (includes vs excludes attributes).

jeromecomte (approved):

PR #650, opened 2026-05-28
SED-4714 step-30-analytics-finalization-before-release

david-stephan:


step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/dashboards/DashboardsGenerator.java
```diff
@@ -181,6 +181,7 @@ public DashboardView createExecutionsOverviewDashboard() {
             .setDashlets(Arrays.asList(
                 new DashboardItem()
                     .setName("Execution duration")
+                    .setId(UUID.randomUUID().toString())
```
jeromecomte, line 184: Not sure if relevant but just to be sure: until now these dashboard items didn't have any Id. Now if we generate UUID upon creation, don't we have the risk that they have each time a new Id? I assume this has no impact as we decide to create them or not based on something else, but jsut to be sure

jeromecomte (dismissed):
Approving with one comment

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/dashboards/DashboardsGenerator.java
david-stephan: @jeromecomte this indeed doesn't impact the dashboard creation logic. Before this change some items had an id (also UUID) and some not, this just alignes all items and prevent some warning in the FE console.

PR #653, opened 2026-05-29
SED-4672 fragments newly added via fragment manager are not registered in automation package yaml

cmisev:


cmisev:
Hi @cl-exense I have only unit tested so far, will play around with IDE for more rigorous test before I leave it for review

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/AutomationPackageReader.java
```diff
@@ -183,32 +181,32 @@ protected AutomationPackageContent newContentInstance() {
...
-            fillAutomationPackageWithImportedFragments(content, descriptor, archive, fragmentMap);
-            StagingAutomationPackageContext stagingContext = new StagingAutomationPackageContext(null, AutomationPackageOperationMode.LOCAL, new LocalResourceManagerImpl(Path.of(descriptorURL.getPath()).getParent().toFile()), archive, content, null, null, new HashMap<>());
-            return new AutomationPackageYamlFragmentManager(descriptor, fragmentMap, getOrCreateDescriptorReader(), stagingContext);
+            Set<AutomationPackageFragmentYaml> fragments = new HashSet<>();
+            fillAutomationPackageWithImportedFragments(content, descriptor, archive, fragments);
+            StagingAutomationPackageContext stagingContext = new StagingAutomationPackageContext(null, AutomationPackageOperationMode.LOCAL, new LocalResourceManagerImpl(Path.of(descriptorUrl.getPath()).getParent().toFile()), archive, content, null, null, new HashMap<>());
+            return new AutomationPackageYamlFragmentManager(descriptor, fragments, getOrCreateDescriptorReader(), stagingContext);
         } catch (IOException e) {
```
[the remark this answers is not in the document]
cmisev: The suggestion will fail where the fragments are loaded directly from a compressed jar. I have solved it in a different way to allow for the option that the fragment path cannot be set

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/AutomationPackageReader.java
```diff
@@ -183,32 +181,32 @@ protected AutomationPackageContent newContentInstance() {
...
-                        fragmentYamlMap.put(resource.toString(), fragment);
-                        fragment.setFragmentUrl(resource);
-                        fillAutomationPackageWithImportedFragments(targetPackage, fragment, archive, fragmentYamlMap);
+                        AutomationPackageFragmentYaml referencedFragment = getOrCreateDescriptorReader().readAutomationPackageFragment(fragmentYamlStream, resource.toString(), archive.getAutomationPackageName());
+                        fragments.add(referencedFragment);
+                        referencedFragment.setFragmentPath(Path.of(resource.getPath()));
+                        fillAutomationPackageWithImportedFragments(targetPackage, referencedFragment, archive, fragments);
                     } catch (IOException e) {
                         throw new AutomationPackageReadingException("Unable to read fragment in automation package: " + importedFragmentReference, e);
                     }
```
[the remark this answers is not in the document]
cmisev: Also here I solved it in a different way.

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
```diff
@@ -222,54 +219,100 @@ private <T extends PatchableYamlModel> void addFragmentEntity(AutomationPackageF
...
+
+
+                    Path referencePath = determineObjectRelativePath(newEntity, fieldName, true);
+                    if (referencingFragment.getFragments().removeIf(f -> f.getValue().equals(oldRelativePath.toString()))) {
+                        referencingFragment.getFragments().add(new PatchableYamlPrimitive<>(referencingFragment.getPatchingContext(), referencePath.toString()));
+                        referencingFragment.writeToDisk();
+                    };
+                } catch (IOException ignored) {
+                }
+            }
```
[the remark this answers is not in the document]
cmisev: Improved error handling

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
```diff
@@ -222,54 +219,100 @@ private <T extends PatchableYamlModel> void addFragmentEntity(AutomationPackageF
...
+        for (AutomationPackageFragmentYaml fragment : fragments) {
+            for (PatchableYamlPrimitive<String> fragmentPathPattern : fragment.getFragments()) {
+                PathMatcher matcher = FileSystems.getDefault().getPathMatcher("glob:" + fragmentPathPattern.getValue());
+                if (matcher.matches(path)) {
+                    return Optional.of(fragment);
+                }
+            }
         }
+        return Optional.empty();
+    }
```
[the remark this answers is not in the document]
cmisev: Replaced glob matching entirely with pre-existing, custom class.

step-core-model/src/main/java/step/core/yaml/deserialization/PatchableYamlPrimitive.java
```diff
@@ -0,0 +1,54 @@
...
+    @JsonCreator(mode = JsonCreator.Mode.DELEGATING)
+    public PatchableYamlPrimitive(@JacksonInject(useInput = OptBoolean.FALSE) PatchingContext context, T value) {
+        super(context);
+        this.value = value;
+    }
+
+    @Override
+    public String toString() {
+        return value.toString();
+    }
```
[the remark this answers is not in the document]
cmisev: Resolved with null check upon creation

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
```diff
@@ -222,54 +219,100 @@ private <T extends PatchableYamlModel> void addFragmentEntity(AutomationPackageF
...
 
-                %2$s = %1$s.yml
-                %3$s = %4$s
-                """, fieldName, String.format(PROPERTY_NEW_OBJECT_FRAGMENT_PATH, fieldName), String.format(PROPERTY_NEW_OBJECT_FRAGMENT_MODE, fieldName), NewObjectFragmentMode.FRAGMENT.name()));
+        Optional<AutomationPackageFragmentYaml> optionalReferencingFragment = determineReferencingFragment(path);
+        if (optionalReferencingFragment.isEmpty()) {
+            Path referencePath = determineObjectRelativePath(p, fieldName, true);
+            descriptorYaml.getFragments().add(new PatchableYamlPrimitive<>(descriptorYaml.getPatchingContext(), referencePath.toString()));
+            descriptorYaml.writeToDisk();
         }
```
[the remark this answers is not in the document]
cmisev: Solved in a different way, encapsulating the path handlling in a pre-existing custom glob treatment class.

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
```diff
@@ -222,54 +224,135 @@ private <T extends PatchableYamlModel> void addFragmentEntity(AutomationPackageF
...
+            String referencingPath = resourcePatchMatchingResolver.getFragmentReferenceString(determineObjectRelativePath(p, fieldName, true));
+            descriptorYaml.getFragments().add(new PatchableYamlPrimitive<>(descriptorYaml.getPatchingContext(), referencingPath));
+            descriptorYaml.writeToDisk();
         }
+        return fragment;
+    }
 
-        if (mode == NewObjectFragmentMode.PER_OBJECT) {
-            path = path.resolve(sanitizeFilename(p.getAttribute(AbstractOrganizableObject.NAME)) + ".yml");
+    private Optional<AutomationPackageFragmentYaml> determineReferencingFragment(Path path) {
```
cl-exense, line 323: This entire method can be rewritten in pure functional style, which makes it both more efficient and (presumably) more understandable:
```
private Optional<AutomationPackageFragmentYaml> determineReferencingFragment(Path path) {
    return Stream.concat(Stream.of(descriptorYaml), fragments.stream())
            .filter(fragment -> fragment.getFragments().stream()
                    .anyMatch(pattern -> resourcePatchMatchingResolver.isMatchingPath(pattern.getValue(), path)))
            .findFirst();
}
```

step-core-model/src/main/java/step/core/dynamicbeans/DynamicValue.java
```diff
@@ -141,4 +141,15 @@ protected DynamicValue<T> _cloneValue(DynamicValue<T> clone) {
     protected boolean hasProtectedAccess() {
         return false;
     }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == null) return false;
+        if (obj instanceof DynamicValue<?> other) {
+            if (dynamic != other.dynamic) return false;
+            if (dynamic) return expression.equals(other.expression);
```
cl-exense: This looks semantically correct, however I'm not sure if expressions can safely be evaluated at any time. If I remember correctly, expressions may require state that could only be present at "runtime" (while a plan is running, such as another variable's content). Example: a dynamic value which resolves the "${ENV}" parameter. This is no problem at runtime -- but ONLY at runtime. At "compile time", the expression simply cannot be resolved yet.

step-core-model/src/main/java/step/core/yaml/PatchingContext.java
```diff
@@ -32,13 +29,16 @@ public PatchingContext() {
...
     public PatchingContext(ObjectMapper mapper) {
-        this(null, "", mapper);
+        this("", "", mapper);
     }
 
     public PatchingContext(String sourceLocation, String yaml, ObjectMapper mapper) {
         this.sourceLocation = sourceLocation;
         this.initialLines = new CopyOnWriteArrayList<>(yaml.lines().toList());
         this.mapper = mapper;
+        Objects.requireNonNull(sourceLocation);
```
cl-exense: These checks are all ok, but should all be inlined above, e.g. this.sourceLocation = Objects.requireNonNull(sourceLocation)

step-core-model/src/main/java/step/core/yaml/PatchingContext.java
```diff
@@ -152,7 +150,7 @@ private List<ChunkBounds> getClaimedOuterBounds() {
     private String serializeUnindented(Object entity) {
         try {
             return mapper.writeValueAsString(entity)
-                .replaceFirst("^---\\s*\\n", "")
+                .replaceFirst("^---\\s*\\n*", "")
```
cl-exense: This drastically changes the logic of the regex! Previously, it would only remove a full line consisting of --- (and optional spaces). Now, it will strip prefixes, i.e. it will also match "---abc" and turn it into "abc". Is this REALLY what we want???

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/AutomationPackageReader.java
```diff
@@ -183,33 +182,46 @@ protected AutomationPackageContent newContentInstance() {
...
+                throw new AutomationPackageReadingException("Failed to read automation package for editing. The most likely cause is that you were trying to load " +
+                    "an automation package as a packaged jar. This is not supported and expected behaviour", e);
+            }
         } catch (IOException e) {
             throw new AutomationPackageReadingException("Failed to read automation package for editing", e);
         }
     }
 
-    private void fillAutomationPackageWithImportedFragments(AutomationPackageContent targetPackage, AutomationPackageFragmentYaml fragment, T archive, Map<String, AutomationPackageFragmentYaml> fragmentYamlMap) throws AutomationPackageReadingException {
+    private void fillAutomationPackageWithImportedFragments(AutomationPackageContent targetPackage, AutomationPackageFragmentYaml fragment, T archive, Set<AutomationPackageFragmentYaml> fragments) throws AutomationPackageReadingException {
```
cl-exense, line 214: This signature is confusing. There is a fragment parameter, and an additional fragments parameter. Please rename the parameters to make clear what exactly they represent.

In fact, after reading it again, I just realized the entire method is confusing... :-) - it modifies the fragments parameter that is passed in, and it also recursively calls itself. So please add some Javadoc explaining the parameters, and what exactly the behavior here is...

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/model/AbstractAutomationPackageFragmentYaml.java
```diff
@@ -146,14 +142,23 @@ public PatchingContext getPatchingContext() {
...
-        } catch (IOException | URISyntaxException e) {
-            throw new AutomationPackageWriteToDiskException(MessageFormat.format("Error when writing automation package fragment {0} back to disk.", url), e);
+        } catch (IOException e) {
+            throw new AutomationPackageWriteToDiskException(MessageFormat.format("Error when writing automation package fragment {0} back to disk.", path), e);
         }
     }
+
+    @Override
+    public boolean isEmpty() {
+        return getFragments().isEmpty() &&
```
cl-exense, line 158: suggestion (note: this is incomplete, based on pre-PR code, so replace with correct values!):
```
        return Stream.of(keywords, plans, plansPlainText, additionalFields.values()).allMatch(Collection::isEmpty);
```

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
```diff
@@ -75,24 +80,24 @@ public enum NewObjectFragmentMode {
 
     protected final Map<AbstractOrganizableObject, PatchableYamlModel> patchableMap = new ConcurrentHashMap<>();
     protected final Map<AbstractOrganizableObject, AutomationPackageFragmentYaml> fragmentMap = new ConcurrentHashMap<>();
```
cl-exense, line 82: There still seems to be some redundancy between the fragmentMap, and the fragments Set??

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
```diff
@@ -222,54 +224,135 @@ private <T extends PatchableYamlModel> void addFragmentEntity(AutomationPackageF
         fragment.writeToDisk();
     }
 
-    private <T extends PatchableYamlModel> void modifyFragmentEntity(AutomationPackageFragmentYaml fragment, PatchableYamlList<T> entityList, T oldEntity, T newEntity) {
+    private <T extends PatchableYamlModel> void modifyFragmentEntity(AutomationPackageFragmentYaml fragment, PatchableYamlList<T> entityList, T oldEntity, T newEntity, String fieldName) {
```
cl-exense, line 227: **Not related to this particular method here, but to the entire concept:**

It's good to try to use the existing constants for the "plans", "keywords", "parameters" etc, but there are multiple annoying problems with that:
1. they're defined all over the place in the code base, in vastly differing classes and with vastly different constant names too - which means they're basically impossible to remember and have to be searched for every time again when trying to write code that uses them.
2. they're simple strings, which makes type checks, autocompletions etc. impossible, and is simply error-prone in general.

Since I've also struggled with this multiple times already, I propose to introduce an enum explicitly for this purpose, something along the lines of

```
public enum AutomationPackageEntity {
    PLAN(YamlPlan.PLANS_ENTITY_NAME),
    KEYWORD(YamlAutomationPackageKeyword.KEYWORDS_ENTITY_NAME);

    public final String listName;

    AutomationPackageEntity(String listName) {
        this.listName = listName;
    }
}
```
Note this is incomplete and subject to potentially finding a better name (and figuring out if singular or plural names are more appropriate). However, it will solve all the abovementioned problems at once (type safety, autocompletion,...), and can be used in most places that are now using simple strings.

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
```diff
@@ -222,54 +224,135 @@ private <T extends PatchableYamlModel> void addFragmentEntity(AutomationPackageF
...
         entityList.replaceItem(oldEntity, newEntity);
+        Path oldRelativePath = determineObjectRelativePath(oldEntity, fieldName, false);
+        Path newRelativePath = determineObjectRelativePath(newEntity, fieldName, false);
+        if (!oldRelativePath.equals(newRelativePath)) {
+            Path absoluteOldPath = apRoot.resolve(oldRelativePath);
+
+            if (absoluteOldPath.equals(fragment.getFragmentPath())) {
```
cl-exense: This may be a stupid question, but do we need an else branch here?

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
```diff
@@ -222,54 +224,135 @@ private <T extends PatchableYamlModel> void addFragmentEntity(AutomationPackageF
         fragment.writeToDisk();
     }
 
-    private <T extends PatchableYamlModel> void modifyFragmentEntity(AutomationPackageFragmentYaml fragment, PatchableYamlList<T> entityList, T oldEntity, T newEntity) {
+    private <T extends PatchableYamlModel> void modifyFragmentEntity(AutomationPackageFragmentYaml fragment, PatchableYamlList<T> entityList, T oldEntity, T newEntity, String fieldName) {
         entityList.replaceItem(oldEntity, newEntity);
+        Path oldRelativePath = determineObjectRelativePath(oldEntity, fieldName, false);
+        Path newRelativePath = determineObjectRelativePath(newEntity, fieldName, false);
+        if (!oldRelativePath.equals(newRelativePath)) {
+            Path absoluteOldPath = apRoot.resolve(oldRelativePath);
```
cl-exense: Naming consistency: you're using oldRelativePath, but absoluteOldPath. Stick to one convention or the other (I prefer the first).

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
```diff
@@ -222,54 +224,135 @@ private <T extends PatchableYamlModel> void addFragmentEntity(AutomationPackageF
...
-        if (mode == NewObjectFragmentMode.FRAGMENT) {
-            defaultRelativeFragmentPath = defaultRelativeFragmentPath + ".yml";
+        PatchableYamlModel yamlObject = patchableMap.get(object);
+        entityList.remove(yamlObject);
+
+        patchableMap.remove(object);
+        fragmentMap.remove(object);
+
+        if (fragment.isEmpty()) {
+            try {
```
cl-exense, line 280: Cleaner and more readable `try` block directly using functional style `ifPresent` instead of temporary variables, `isPresent` and `get`:

```
FileUtils.delete(fragment.getFragmentPath().toFile());

determineReferencingFragment(fragment.getFragmentPath()).ifPresent(referencingFragment -> {
    String relativeFragmentReference = resourcePatchMatchingResolver
            .getFragmentReferenceString(apRoot.relativize(fragment.getFragmentPath()));
            
    if (referencingFragment.getFragments().removeIf(f -> f.getValue().equals(relativeFragmentReference))) {
        referencingFragment.writeToDisk();
    }

    fragments.remove(fragment);
});
```

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
```diff
@@ -222,54 +224,135 @@ private <T extends PatchableYamlModel> void addFragmentEntity(AutomationPackageF
...
-        if (!path.isAbsolute()) {
-            path = apRoot.resolve(path);
+        PatchingContext context = new PatchingContext(absolutePath.toString(), "---", descriptorYaml.getPatchingContext().getMapper());
+        AutomationPackageFragmentYaml fragment = new AutomationPackageFragmentYamlImpl(context);
+        fragments.add(fragment);
+        fragment.setFragmentPath(absolutePath);
+
+
+        Optional<AutomationPackageFragmentYaml> optionalReferencingFragment = determineReferencingFragment(path);
+        if (optionalReferencingFragment.isEmpty()) {
```
cl-exense: `if (determineReferencingFragment(path).isEmpty()) {` -- the variable is unnecessary.

cl-exense (changes requested):
Generally looks ok, just a few questions and suggestions for improvements .

cl-exense:
@cmisev We have the additional issue that this branch and the one I worked on (SED-4557) have slightly diverged. Not terribly much, but there will be a few conflicts. I suggest to address the review comments first, then we can see how to consolidate. The 4557 is still ongoing, so if this here can then be considered finished, it makes sense to merge it to 4427 (and then into 4557, fixing conflicts).

cmisev:
Hi @cl-exense thanks for the comments, will adress them so we can quickly move on. I am not too much worried about potential divergence with SED-4557 but rather between SED-4557 and SED-4770, would you mind creating a draft PR for SED-4557 so I can see your current state?

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
cmisev: Thanks, suggestion taken as is.

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
cmisev: @cl-exense I was looking for something like that. I am suggesting though to implement this in https://github.com/exense/step/pull/659 as this new PR is a lot about streamlining and simplifying.

One issue i see with enums is that it does not work well with extending support for additional entities in a plug-in fashion, and the (shared) enum somehow already needs to know about schedules for example.

In https://github.com/exense/step/pull/659 I solved this by having a overridable method "getCollectionName()" on the object to yaml object mappers, maybe this would be sufficient to abstract it away?

step-core-model/src/main/java/step/core/dynamicbeans/DynamicValue.java
cmisev: Seems like this is not needed any more. I implemented it for full support of

`@JsonInclude(JsonInclude.Include.NON_DEFAULT)`

in some entity, however I might have solved the issues related to it in a redundant way. So I remove it again, because equals implementations are often a landmine waiting to go off...

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
cmisev: Suggestion taken

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
cmisev: removed

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
cmisev: Done, Thanks

step-core-model/src/main/java/step/core/yaml/PatchingContext.java
cmisev: Your example is the one I needed it for - base types such as strings were written as 

`--- "fragment.yml"`

without the correction.

HOWEVER: I got rid of the whole transformation by simply disabling the yaml generator feature which prints the "---" prefix.

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
cmisev: Not a stupid question. I have added a comment for clarity

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/AutomationPackageReader.java
cmisev: Javadoc added

cmisev:
@cl-exense Just saw your draft https://github.com/exense/step/pull/667/changes#diff-3a67b8eed37d9e216d5c4bc4c2e11223a0a0481967d94bf40ca862e4f99d69eb , am not all to worried about divergence for now.

PR #659, opened 2026-06-19
SED-4770 handle more keyword types

cmisev:


cmisev:
Will add some unit tests for jmeter and node keywords

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/model/AbstractAutomationPackageFragmentYaml.java
```diff
@@ -161,4 +165,25 @@ public boolean isEmpty() {
...
+    public <YO extends PatchableYamlModel, BO extends AbstractOrganizableObject> PatchableYamlList<YO> getListForYamlObject(HasCollectionName<BO, YO> mapper) {
+        String collectionName = mapper.getCollectionName();
+        return (PatchableYamlList<YO>) switch (collectionName) {
+            case YamlAutomationPackageKeyword.KEYWORDS_ENTITY_NAME -> getKeywords();
+            case YamlPlan.PLANS_ENTITY_NAME -> getPlans();
+            case "plansPlainText" -> getPlansPlainText();
+            default -> getAdditionalFields()
+                .computeIfAbsent(collectionName, n -> new PatchableYamlList<YO>(getPatchingContext(), n));
+        };
+    }
```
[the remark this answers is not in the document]
cmisev: made getPlainTextPlans return patchableyamllist, and generally referenced the fields directly

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/mappers/CompositeFunctionToYamlMapper.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Override
+    public YamlAutomationPackageKeyword getNewYamlObject(CompositeFunction compositeFunction) {
+
+        YamlCompositeFunction yamlComposite = new YamlCompositeFunction();
+        setCommonAtributes(compositeFunction, yamlComposite);
+
+        YamlPlan plan = planReader.planToYamlPlan(compositeFunction.getPlan());
+        plan.setName(null);
+        yamlComposite.setPlan(plan);
```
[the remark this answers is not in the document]
cmisev: Improved in a different way

cmisev:
@cl-exense I now had to implement the equals for dynamic values again.... I think it is due to serialization/deserialization over the ide api services that the new instances are created and then comparison by object is not enough. let's discuss whether you think the new implementation is better...

step-ap-ide/src/main/resources/work-initial/jmeterProject1/jmeterProject1.xml
cl-exense, line 1: The entire step-ap-ide module has been removed, please delete the corresponding files.

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/AutomationPackageReader.java
```diff
@@ -182,33 +182,56 @@ protected AutomationPackageContent newContentInstance() {
...
-                        fragmentYamlMap.put(resource.toString(), fragment);
-                        fragment.setFragmentUrl(resource);
-                        fillAutomationPackageWithImportedFragments(targetPackage, fragment, archive, fragmentYamlMap);
-                    } catch (IOException e) {
+                        AutomationPackageFragmentYaml referencedFragment = getOrCreateDescriptorReader().readAutomationPackageFragment(fragmentYamlStream, resource.toString(), archive.getAutomationPackageName());
+                        fragments.add(referencedFragment);
+                        try {
+                            referencedFragment.setFragmentPath(Path.of(resource.toURI()));
+                        } catch (FileSystemNotFoundException e) {
+                            logger.warn("Could not set Fragment path for fragment editing while loading fragment. " +
```
cl-exense, line 234: In very similar code just above we're throwing an exception. Here, we're just logging a warning. Is this intentional? Note: users likely won't see the warning anyway, so is it really safe to just ignore these errors? In addition, you're already modifying the `fragments` Set before this operation which can potentially throw an exception, meaning that if it gets thrown, you'll have broken content in the `fragments` (I suppose).

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/deserialization/AbstractYamlAutomationPackageFragmentDeserializer.java
```diff
@@ -27,25 +27,19 @@
...
     public AbstractYamlAutomationPackageFragmentDeserializer(BeanDeserializer deserializer) {
         super(deserializer);
-        delegate = deserializer;
     }
 
     @Override
-    public Object deserialize(JsonParser p, DeserializationContext ctxt) throws IOException {
-        return deserialize(p, ctxt, new AutomationPackageFragmentYamlImpl((PatchingContext) ctxt.getAttribute(PatchingContext.class)));
-    }
+    abstract public Object deserialize(JsonParser p, DeserializationContext ctxt) throws IOException;
```
cl-exense: Coding convention: `public abstract`.

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/mappers/CompositeFunctionToYamlMapper.java
```diff
@@ -0,0 +1,54 @@
...
+import step.automation.packages.mappers.AbstractFunctionToYamlMapper;
+import step.automation.packages.mappers.interfaces.BusinessObjectToYamlMapping;
+import step.automation.packages.model.YamlAutomationPackageKeyword;
+import step.core.plans.Plan;
+import step.plans.parser.yaml.YamlPlan;
+import step.plans.parser.yaml.YamlPlanReader;
+import step.plugins.functions.types.CompositeFunction;
+import step.plugins.functions.types.automation.YamlCompositeFunction;
+
+/*******************************************************************************
```
cl-exense, line 1: I'm not a fan of these copyright headers anyway, but I'm seemingly alone with that opinion :) . Anyway, if you insert it, put it at the very beginning of the file, before anything else, not in the middle of the file...

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/mappers/GeneralScriptFunctionToYamlMapper.java
```diff
@@ -0,0 +1,44 @@
+package step.automation.packages.yaml.mappers;
+
+import step.automation.packages.mappers.AbstractFunctionToYamlMapper;
+import step.automation.packages.model.YamlAutomationPackageKeyword;
+import step.automation.packages.mappers.interfaces.BusinessObjectToYamlMapping;
+import step.plugins.java.GeneralFunctionScriptLanguage;
+import step.plugins.java.GeneralScriptFunction;
+import step.plugins.java.automation.YamlGeneralScriptFunction;
+
+/*******************************************************************************
```
cl-exense, line 1: Idem.

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/mappers/GeneralScriptFunctionToYamlMapper.java
```diff
@@ -0,0 +1,44 @@
...
+
+    @Override
+    public YamlAutomationPackageKeyword getNewYamlObject(GeneralScriptFunction generalScriptFunction) {
+
+        YamlGeneralScriptFunction yamlFunction = new YamlGeneralScriptFunction();
+        setCommonAttributes(generalScriptFunction, yamlFunction);
+
+        yamlFunction.setScriptFile(generalScriptFunction.getScriptFile());
+        yamlFunction.setLibrariesFile(generalScriptFunction.getLibrariesFile());
+        yamlFunction.setScriptLanguage(GeneralFunctionScriptLanguage.valueOf(generalScriptFunction.getScriptLanguage().getValue()));
```
cl-exense, line 39: Can getScriptLanguage() return null? -> potential NPE.

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/mappers/ParameterMapper.java
```diff
@@ -0,0 +1,49 @@
+package step.automation.packages.yaml.mappers;
+
+import step.automation.packages.mappers.interfaces.BusinessObjectToYamlMapper;
+import step.automation.packages.mappers.interfaces.BusinessObjectToYamlMapping;
+import step.automation.packages.mappers.interfaces.YamlToBusinessObjectMapper;
+import step.automation.packages.mappers.interfaces.YamlToBusinessObjectMapping;
+import step.parameter.Parameter;
+import step.parameter.automation.AutomationPackageParameter;
+
+/*******************************************************************************
```
cl-exense, line 1: Idem.

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/mappers/YamlToFunctionMapper.java
```diff
@@ -0,0 +1,45 @@
+package step.automation.packages.yaml.mappers;
+
+import step.automation.packages.StagingAutomationPackageContext;
+import step.automation.packages.model.YamlAutomationPackageKeyword;
+import step.automation.packages.mappers.interfaces.YamlToBusinessObjectMapper;
+import step.automation.packages.mappers.interfaces.YamlToBusinessObjectMapping;
+import step.functions.Function;
+
+/*******************************************************************************
```
cl-exense, line 1: Idem.

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/mappers/PlanMapper.java
```diff
@@ -0,0 +1,56 @@
...
+
+import step.automation.packages.mappers.interfaces.BusinessObjectToYamlMapper;
+import step.automation.packages.mappers.interfaces.BusinessObjectToYamlMapping;
+import step.automation.packages.mappers.interfaces.YamlToBusinessObjectMapper;
+import step.automation.packages.mappers.interfaces.YamlToBusinessObjectMapping;
+import step.core.plans.Plan;
+import step.plans.parser.yaml.YamlPlan;
+import step.plans.parser.yaml.YamlPlanReader;
+
+/*******************************************************************************
```
cl-exense, line 1: Idem.

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/model/AbstractAutomationPackageFragmentYaml.java
```diff
@@ -18,34 +18,33 @@
  ******************************************************************************/
 package step.automation.packages.yaml.model;
 
-import com.fasterxml.jackson.annotation.JsonAnyGetter;
-import com.fasterxml.jackson.annotation.JsonIgnore;
-import com.fasterxml.jackson.annotation.JsonInclude;
-import com.fasterxml.jackson.annotation.JsonSetter;
-import com.fasterxml.jackson.annotation.Nulls;
+import com.fasterxml.jackson.annotation.*;
```
cl-exense: Please avoid wildcard imports in the future (But I guess we already changed the configuration accordingly).

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
```diff
@@ -18,45 +18,39 @@
...
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public class AutomationPackageYamlFragmentManager {
 
-
     protected final Path apRoot;
-    protected final StagingAutomationPackageContext stagingContext;
+    private final ResourcePathMatchingResolver resourcePatchMatchingResolver;
```
cl-exense: Typo: resourcePat**c**h ...

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
```diff
@@ -71,274 +65,255 @@ public enum NewObjectFragmentMode {
...
 
-    public void setProperties(Properties properties) {
-        this.properties = properties;
+    private void scanBusinessObjectMappers(Map<Class<?>, Object> injectables) {
+
+        for (Class<?> annotatedClass : CachedAnnotationScanner.getClassesWithAnnotation(BusinessObjectToYamlMapping.class)) {
+            Arrays.stream(annotatedClass.getConstructors()).findFirst().ifPresent(c -> {
+                try {
+                    BusinessObjectToYamlMapping annotation = annotatedClass.getAnnotation(BusinessObjectToYamlMapping.class);
+                    mappers.put(annotation.sourceClass(), (BusinessObjectToYamlMapper<?, ?>) c.newInstance(Arrays.stream(c.getParameterTypes()).map(injectables::get).toArray()));
```
cl-exense: This gets really heavy to read.
Proposed change to increase understandability a bit -- introduce intermediate var (and rename some others):

`var mapper = (BusinessObjectToYamlMapper<?, ?>) constructor.newInstance(Arrays.stream(constructor.getParameterTypes()).map(injectables::get).toArray());`

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
```diff
@@ -71,274 +65,255 @@ public enum NewObjectFragmentMode {
...
-        apRoot = Path.of(descriptorYaml.getFragmentUrl().getPath())
-            .getParent();
+        apRoot = descriptorYaml.getFragmentPath().getParent();
+
+        Map<Class<?>, Object> injectables = new HashMap<>();
+
+        injectables.put(YamlPlanReader.class, descriptorReader.getPlanReader());
+        injectables.put(StagingAutomationPackageContext.class, stagingContext);
+
+        scanBusinessObjectMappers(injectables);
```
cl-exense: 1. inconsistent initialization and naming: the one direction uses a pre-initialized final field named `mappers`, the other direction creates and returns an object which is stored as a local variable named `yamlObjectMappers`. Not sure if the latter also needs a field, but I suggest to at least make the intentions/naming clearer, e.g. `yaml2boMappers`, `bo2yamlMappers` or so.
2. Propose to rename the methods to "create..." instead of "scan...", and have both methods return the object they created.

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
```diff
@@ -71,274 +65,255 @@ public enum NewObjectFragmentMode {
...
-            .forEach(this::initializeMaps);
+            .forEach(f -> initializeMaps(f, yamlObjectMappers));
     }
 
-    public void setProperties(Properties properties) {
-        this.properties = properties;
+    private void scanBusinessObjectMappers(Map<Class<?>, Object> injectables) {
+
+        for (Class<?> annotatedClass : CachedAnnotationScanner.getClassesWithAnnotation(BusinessObjectToYamlMapping.class)) {
+            Arrays.stream(annotatedClass.getConstructors()).findFirst().ifPresent(c -> {
```
cl-exense: This is ok implementation-wise, but it imposes an important restriction, namely that effectively every class carrying this annotation must have a single constructor, and that all constructor arguments can be "satisfied" from the injectables here. This should be documented somewhere (presumably in the Mapping annotation.). 

I also have the feeling that there is plenty that could potentially go wrong here, so add null checks (potentially already earlier, e.g. in constructor), and some logging in case of unexpected situations. Note: Once the code is stable usually nothing should go wrong, but this is a landmine ready to explode during development, and it will be hard to debug without proper error handling.

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
```diff
@@ -71,274 +65,255 @@ public enum NewObjectFragmentMode {
...
-    public void initializeMaps(AutomationPackageFragmentYaml fragment) {
-        pathToYamlFragment.put(fragment.getFragmentUrl().toString(), fragment);
-        for (YamlPlan yamlPlan : fragment.getPlans()) {
-            Plan plan = descriptorReader.getPlanReader().yamlPlanToPlan(yamlPlan);
-            patchableMap.put(plan, yamlPlan);
-            fragmentMap.put(plan, fragment);
+    private Collection<YamlToBusinessObjectMapper<?, ?>> scanYamlObjectMappers(Map<Class<?>, Object> injectables) {
+        List<YamlToBusinessObjectMapper<?, ?>> list = new LinkedList<>();
+        for (Class<?> annotatedClass : CachedAnnotationScanner.getClassesWithAnnotation(YamlToBusinessObjectMapping.class)) {
+            Arrays.stream(annotatedClass.getConstructors()).findFirst().ifPresent(c -> {
```
cl-exense: Similar comment applies as for the other method above.

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
```diff
@@ -71,274 +65,255 @@ public enum NewObjectFragmentMode {
 
     public static final String PROPERTY_NEW_OBJECT_FRAGMENT_MODE = "newFragmentPaths.%s.mode";
     public static final String PROPERTY_NEW_OBJECT_FRAGMENT_PATH = "newFragmentPaths.%s.path";
-    protected final AutomationPackageDescriptorReader descriptorReader;
 
     protected final Map<AbstractOrganizableObject, PatchableYamlModel> patchableMap = new ConcurrentHashMap<>();
```
cl-exense, line 82: I'm still not very happy with keeping multiple collections (patchableMap, fragmentMap, and fragments) in parallel, because they also all need to be maintained in parallel, and care must be taken to keep everything in sync so that they don't diverge.

 If I understand correctly:
1. Every `AbstractOrganizableObject` should ALWAYS map both to a `PatchableYamlModel` and to a `AutomationPackageFragmentYaml`, right? Therefore, the first two might be combined into something like a `Map<AOO, Yamls> objectToYamls`, where the value is simply `record Yamls(PatchableYamlModel model, AutomationPackageYamlFragment fragment)`. This should get rid of the first redundancy (the naming could probably be improved, but you get the idea).
2. The `fragments` set does not actually have to be maintained separately at all. Yes, it is passed in in the constructor where it's iterated over to construct the other map(s). However after that, the actual field is redundant as the only place where it's actually read is when trying to find the fragment for a path. At that point, fragments can easily be determined by iterating over the (unique) fragment values of the existing map from above. This should drastically simplify the code. What do you think?

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/model/AutomationPackageFragmentYaml.java
```diff
@@ -47,13 +52,19 @@ default <T> PatchableYamlList<T> getAdditionalField(String k) {
...
 
     void setPatchingContext(PatchingContext context);
 
     void writeToDisk();
+
+    boolean isEmpty();
+
+    <YO extends PatchableYamlModel, BO extends AbstractOrganizableObject> void initializeMaps(YamlToBusinessObjectMapper<BO, YO> mapper, Map<AbstractOrganizableObject, PatchableYamlModel> patchableMap, Map<AbstractOrganizableObject, AutomationPackageFragmentYaml> fragmentMap);
+
+    <YO extends PatchableYamlModel, BO extends AbstractOrganizableObject> PatchableYamlList<YO> getListForYamlObject(HasCollectionName<BO,YO> mapper);
```
cl-exense: This is an **extremely** confusing method signature (and usage), because  the method name, argument name and argument type seemingly do not match at all.

In fact, the (only) implementation is only using the collectionName, so the signature should probably simply be getListForCollection(String collectionName), and be called accordingly. This in turn also allows to get rid of the HasCollectionName interface which has weird type arguments :-D -- unless you envisage it for some other future use.

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
```diff
@@ -71,274 +65,255 @@ public enum NewObjectFragmentMode {
...
+
+                try {
+                    FileUtils.moveFile(oldAbsolutePath.toFile(), newAbsolutePath.toFile());
+                    fragment.setFragmentPath(newAbsolutePath);
+                } catch (IOException e) {
+                    throw new AutomationPackageConcurrentEditException(
+                        String.format("Unable to rename file %s to file %s. Was the file renamed or deleted outside the editor?", oldAbsolutePath, newAbsolutePath));
+                }
+
+                AutomationPackageFragmentYaml referencingFragment = determineReferencingFragment(oldRelativePath)
```
cl-exense, line 240: Add another comment explaining the logic below, as it's not easy to understand (especially the final if block)

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
```diff
@@ -71,274 +65,255 @@ public enum NewObjectFragmentMode {
...
+        } else {
+            fragment.writeToDisk();
         }
-
     }
 
-    public String sanitizeFilename(String inputName) {
-        return URLEncoder.encode(inputName, Charset.defaultCharset()).replace("+", " ");
-    }
+    private AutomationPackageFragmentYaml fragmentForNewObject(PatchableYamlModel p, String fieldName) {
```
cl-exense: `p` is a poor name for a parameter...

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
```diff
@@ -71,274 +65,255 @@ public enum NewObjectFragmentMode {
...
 
-    public void removePlan(Plan plan) {
-        AutomationPackageFragmentYaml fragment = fragmentMap.get(plan);
-        YamlPlan yamlPlan = (YamlPlan) patchableMap.get(plan);
+        Path path = determineObjectRelativePath(p, fieldName, false);
```
cl-exense: `relativeFragmentPath` or similar..

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
```diff
@@ -71,274 +65,255 @@ public enum NewObjectFragmentMode {
...
 
-        patchableMap.remove(plan);
-        fragmentMap.remove(plan);
+        PatchingContext context = new PatchingContext(absolutePath.toString(), "---", descriptorYaml.getPatchingContext().getMapper());
+        AutomationPackageFragmentYaml fragment = new AutomationPackageFragmentYamlImpl(context);
+        fragments.add(fragment);
+        fragment.setFragmentPath(absolutePath);
 
-        fragment.writeToDisk();
+        if (determineReferencingFragment(path).isEmpty()) {
```
cl-exense, line 299: Add a comment what this code does/why it's needed

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
```diff
@@ -71,274 +65,255 @@ public enum NewObjectFragmentMode {
...
-
-        fragment.writeToDisk();
+    private Optional<AutomationPackageFragmentYaml> determineReferencingFragment(Path path) {
+        return Stream.concat(Stream.of(descriptorYaml), fragments.stream())
+            .filter(fragment -> fragment.getFragments().stream()
+                .anyMatch(pattern -> resourcePatchMatchingResolver.isMatchingPath(pattern.getValue(), path)))
+            .findFirst();
     }
 
+    private Path determineObjectRelativePath(PatchableYamlModel p, String fieldName, boolean globPattern) {
```
cl-exense: add some comment or javadoc explaining the behavior and arguments

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
```diff
@@ -71,274 +65,255 @@ public enum NewObjectFragmentMode {
...
-        return object;
+                    %2$s = %1$s.yml
+                    %3$s = %4$s
+                    """, fieldName, String.format(PROPERTY_NEW_OBJECT_FRAGMENT_PATH, fieldName), String.format(PROPERTY_NEW_OBJECT_FRAGMENT_MODE, fieldName), NewObjectFragmentMode.FRAGMENT.name()));
+            }
+        };
     }
 
-
+    private String sanitizeFilename(String inputName) {
```
cl-exense, line 352: AI isn't terribly happy with this, and I agree that it won't produce very pretty file names, but I guess it's usable. Still, use StandardCharsets.UTF_8 instead of some potentially weird local charset, to guarantee UTF8 encoding.

step-core-model/src/main/java/step/core/yaml/deserialization/PatchableYamlPrimitive.java
```diff
@@ -0,0 +1,57 @@
+package step.core.yaml.deserialization;
+
+import com.fasterxml.jackson.annotation.*;
+import com.fasterxml.jackson.core.JsonLocation;
+import step.core.yaml.PatchableYamlModelBase;
+import step.core.yaml.PatchingContext;
+
+import java.util.Objects;
+
+/*******************************************************************************
```
cl-exense, line 1: :-)

step-core-model/src/main/java/step/core/yaml/deserialization/PatchableYamlPrimitive.java
```diff
@@ -0,0 +1,57 @@
...
+        return value;
+    }
+
+    public void setValue(T value) {
+        this.value = value;
+    }
+
+    @Override
+    public void onParsed(JsonLocation startLocation, JsonLocation endLocation) {
+        getPatchingContext().claimChunk(startLocation, startLocation, this);
```
cl-exense, line 60: Sure that both arguments should be startLocation?

step-core-model/src/main/java/step/core/yaml/NamedObjectPatchableYamlModel.java
```diff
@@ -0,0 +1,24 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with STEP.  If not, see <http://www.gnu.org/licenses/>.
+ ******************************************************************************/
+package step.core.yaml;
+
+public interface NamedObjectPatchableYamlModel {
```
cl-exense: Not sure if "Named**Object**" is required, I personally find `NamedPatchableYamlModel` sufficient and more elegant. In any case, please make it `extends PatchableYamlModel`.

step-core/src/main/java/step/attachments/FileResolver.java
```diff
@@ -33,6 +34,8 @@ public class FileResolver {
     public static final String RESOURCE_PREFIX = "resource:";
     public static final String RESOURCE_PATH_SEPARATOR = ":";
 
+    private Path plainPathRoot = Path.of("");
```
cl-exense: Hmmm... "plainPathRoot" sounds kinda weird, how about "rawRootPath"?

step-core/src/main/java/step/attachments/FileResolver.java
```diff
@@ -52,7 +59,7 @@ public File resolve(String path) {
         } else if (path.startsWith(RESOURCE_PREFIX)) {
             file = getResourceRevisionFileHandleForPath(path).getResourceFile();
         } else {
-            file = new File(path);
+            file = plainPathRoot.resolve(Path.of(path)).toFile();
```
cl-exense: This will throw an NPE if the setter wasn't invoked before. It's probably ok to throw, but still guard it with an Objects.requireNonNull("... is not set") or so to make diagnosing easier.

step-core/src/main/java/step/automation/packages/mappers/interfaces/BusinessObjectToYamlMapper.java
```diff
@@ -0,0 +1,27 @@
+package step.automation.packages.mappers.interfaces;
+
+import step.core.accessors.AbstractOrganizableObject;
+import step.core.yaml.PatchableYamlModel;
+
+/*******************************************************************************
```
cl-exense, line 1: :-)

step-core/src/main/java/step/automation/packages/mappers/interfaces/BusinessObjectToYamlMapping.java
```diff
@@ -0,0 +1,33 @@
+package step.automation.packages.mappers.interfaces;
+
+import step.core.accessors.AbstractOrganizableObject;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+/*******************************************************************************
```
cl-exense, line 1: :-)

step-core/src/main/java/step/automation/packages/mappers/interfaces/HasCollectionName.java
```diff
@@ -0,0 +1,27 @@
+package step.automation.packages.mappers.interfaces;
+
+import step.core.accessors.AbstractOrganizableObject;
+import step.core.yaml.PatchableYamlModel;
+
+/*******************************************************************************
```
cl-exense: :-)

step-core/src/main/java/step/automation/packages/mappers/interfaces/YamlToBusinessObjectMapper.java
```diff
@@ -0,0 +1,27 @@
+package step.automation.packages.mappers.interfaces;
+
+import step.core.accessors.AbstractOrganizableObject;
+import step.core.yaml.PatchableYamlModel;
+
+/*******************************************************************************
```
cl-exense, line 1: :-)

step-core/src/main/java/step/automation/packages/mappers/interfaces/HasCollectionName.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * STEP is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with STEP.  If not, see <http://www.gnu.org/licenses/>.
+ ******************************************************************************/
+public interface HasCollectionName<BO extends AbstractOrganizableObject, YO extends PatchableYamlModel> {
```
cl-exense: Note that I believe that this class is superfluous (see some previous comment)

step-core/src/main/java/step/automation/packages/mappers/interfaces/YamlToBusinessObjectMapping.java
```diff
@@ -0,0 +1,33 @@
+package step.automation.packages.mappers.interfaces;
+
+import step.core.yaml.PatchableYamlModel;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+/*******************************************************************************
```
cl-exense, line 1: :-P

step-core/src/main/java/step/automation/packages/mappers/AbstractFunctionToYamlMapper.java
```diff
@@ -0,0 +1,40 @@
...
+
+import step.automation.packages.model.AbstractYamlFunction;
+import step.automation.packages.model.YamlAutomationPackageKeyword;
+import step.automation.packages.mappers.interfaces.BusinessObjectToYamlMapper;
+import step.core.accessors.AbstractOrganizableObject;
+import step.functions.Function;
+
+import java.util.Optional;
+
+/*******************************************************************************
```
cl-exense, line 1: ...

step-core/src/main/java/step/automation/packages/model/YamlAutomationPackageKeyword.java
```diff
@@ -19,14 +19,19 @@
 package step.automation.packages.model;
 
 import step.automation.packages.StagingAutomationPackageContext;
+import step.core.yaml.NamedObjectPatchableYamlModel;
 import step.core.yaml.PatchableYamlModelBase;
 import step.core.yaml.PatchingContext;
 import step.functions.Function;
 
-public class YamlAutomationPackageKeyword extends PatchableYamlModelBase implements AutomationPackageKeyword {
+public class YamlAutomationPackageKeyword extends PatchableYamlModelBase
```
cl-exense: No need for a linebreak before `implements`

step-core/src/main/java/step/automation/packages/model/YamlAutomationPackageKeyword.java
```diff
@@ -19,14 +19,19 @@
...
+    implements AutomationPackageKeyword, NamedObjectPatchableYamlModel {
 
-    private AbstractYamlFunction<?> yamlKeyword;
+    private final AbstractYamlFunction<?> yamlKeyword;
 
+    public YamlAutomationPackageKeyword(AbstractYamlFunction<?> yamlKeyword) {
```
cl-exense: Minor nitpick: there are many implementations calling the constructor with an explicit `null` context, and only one calling this newly created one. I in fact prefer the explicit ones where it's clear that it's explicitly null (I double-checked, and all code paths should actually set it to a valid value later, so that's ok).

step-core/src/main/java/step/automation/packages/AutomationPackageStatus.java
```diff
@@ -21,5 +21,5 @@
 public enum AutomationPackageStatus {
     DELAYED_UPDATE,
     SCHEDULED_RELOAD,
-    RELOAD_FAILED
+    EDIT, RELOAD_FAILED
```
cl-exense: EDITING ?

step-core/src/main/java/step/automation/packages/AutomationPackageLocalResourceMapper.java
```diff
@@ -0,0 +1,39 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with STEP.  If not, see <http://www.gnu.org/licenses/>.
+ ******************************************************************************/
+package step.automation.packages;
+
+import java.nio.file.Path;
+
+public class AutomationPackageLocalResourceMapper extends AutomationPackageResourceUploader {
```
cl-exense, line 26: Note sure what these classes are really there for, but the naming seems inconsistent.

step-functions-plugins/step-functions-plugins-jmeter/step-functions-plugins-jmeter-def/src/main/java/step/plugins/jmeter/automation/JMeterFunctionToYamlMapper.java
```diff
@@ -0,0 +1,38 @@
+package step.plugins.jmeter.automation;
+
+import step.automation.packages.mappers.AbstractFunctionToYamlMapper;
+import step.automation.packages.mappers.interfaces.BusinessObjectToYamlMapping;
+import step.automation.packages.model.YamlAutomationPackageKeyword;
+import step.plugins.jmeter.JMeterFunction;
+
+/*******************************************************************************
```
cl-exense, line 1: ...

step-functions-plugins/step-functions-plugins-jmeter/step-functions-plugins-jmeter-def/src/main/java/step/plugins/jmeter/automation/JMeterFunctionToYamlMapper.java
```diff
@@ -0,0 +1,38 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with STEP.  If not, see <http://www.gnu.org/licenses/>.
+ ******************************************************************************/
+@BusinessObjectToYamlMapping(sourceClass = JMeterFunction.class)
+public class JMeterFunctionToYamlMapper extends AbstractFunctionToYamlMapper<JMeterFunction> {
+
+    @Override
+    public YamlAutomationPackageKeyword getNewYamlObject(JMeterFunction jmeterFunction) {
```
cl-exense: Just realized now: "`getNew`..." is weird naming for a method, the standard naming would be "create..." or so (get usually implies "getter" semantics). In fact, to keep the names as neutral as possible, one could use `toYamlObject` (and `toBusinessObject` respectively in the other mapper class).

step-functions-plugins/step-functions-plugins-node/step-functions-plugins-node-def/src/main/java/step/plugins/node/automation/NodeFunctionToYamlMapper.java
```diff
@@ -0,0 +1,39 @@
+package step.plugins.node.automation;
+
+import step.automation.packages.mappers.AbstractFunctionToYamlMapper;
+import step.automation.packages.mappers.interfaces.BusinessObjectToYamlMapping;
+import step.automation.packages.model.YamlAutomationPackageKeyword;
+import step.plugins.node.NodeFunction;
+
+/*******************************************************************************
```
cl-exense, line 1: ...

step-ide/src/main/java/step/ide/LocalIDEControllerPlugin.java
```diff
@@ -8,15 +9,23 @@
...
     public void serverStart(GlobalContext context) throws Exception {
         System.out.println(this + " serverStart");
         LocalIDEState.get().setResourceManager((ResourceManagerImpl) context.getResourceManager());
+        LocalIDEState.get().setFileResolver(context.getFileResolver());
         var services = context.getServiceRegistrationCallback();
         services.registerService(LocalIDEServices.class);
         services.registerService(LocalFileSystemServices.class);
         context.put(ExecutionDiversion.class, LocalIDEState.get());
     }
+    /*
```
cl-exense: Please remove :)

cl-exense (changes requested):
Uff... Thanks! There are quite a few comments, but the majority (about 2/3 I think) are rather minor. Still, some may require a bit deeper thinking or discussion.

step-core/src/main/java/step/automation/packages/ResourcePathMatchingResolver.java
```diff
@@ -109,4 +114,15 @@ private void findPathMatchingResourcesRecursive(String[] pathArray, int currentL
         }
     }
 
+    private Pattern prepareMatchPattern(String referencingString) {
+        return Pattern.compile(referencingString.replace("*", ".*"));
```
jeromecomte: We should probably quote the referencingString using Pattern.quote

step-functions-plugins/step-functions-plugins-jmeter/step-functions-plugins-jmeter-def/src/main/java/step/plugins/jmeter/automation/JMeterFunctionToYamlMapper.java
jeromecomte: Agree

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
```diff
@@ -71,274 +65,255 @@ public enum NewObjectFragmentMode {
...
-            if (newKeyword != null) {
-                patchableMap.put(function, newKeyword);
-                addFragmentEntity(fragment, fragment.getKeywords(), newKeyword);
-            } else {
-                System.err.println("SAVING OF FUNCTION OF TYPE " + function.getClass().getName() + " IS NOT CURRENTLY SUPPORTED");
-            }
+            fragment = fragmentForNewObject(newYamlObject, mapper.getCollectionName());
+            fragmentMap.put(object, fragment);
+            newYamlObject.setPatchingContext(fragment.getPatchingContext());
+            addFragmentEntity(fragment, fragment.getListForYamlObject(mapper), newYamlObject);
```
jeromecomte: I assume that the main descriptor is considered as fragment and this path (fragment = null) corresponds to the path of a new entity

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
jeromecomte: Agree. We should explicitly search for the constructor with the expected signature and not simply get the first one although we would catch such errors at the latest in the JUnit tests

jeromecomte (approved):
Approving with some minor additional remarks to the review of @cl-exense. None of the feedbacks are critical and should block a merge to the branch SED-4557

step-core/src/main/java/step/attachments/FileResolver.java
cl-exense: Comment was wrong, actually the value is set with a default of "".

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/AutomationPackageReader.java
cl-exense: Clarified for now we keep this implementation, might revise later.

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
cl-exense: Clarified: yes, exactly. APDescriptorYaml extends APFragmentYaml, and the only way case when fragmentMap.get(...) returns null is when saving newly created objects

step-core-model/src/main/java/step/core/yaml/deserialization/PatchableYamlPrimitive.java
cl-exense: Sigh. There actually seems to be a bug in the JSON/YAML parser implementation leading up to here. In most cases, the provided endLocation is reported to be **BEFORE** the startLocation -- fortunately this only seems to happen to single-line primitive strings (so far), so this implementation here kind of "mitigates" the bug -- but still it's clearly a bug.

PR #660, opened 2026-06-22
SED-4758 Implement Label Value Cardinality Safeguards for Custom Metrics Ingestion

david-stephan:


step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesMetricSamplesHandler.java
```diff
@@ -146,9 +220,88 @@ private BucketAttributes metricSampleToBucketAttributes(StepMetricSample mm) {
...
+    /**
+     * Atomically records the value for the given (execution, metric, label) tuple and decides whether it
+     * passes through or must be masked:
+     * <ul>
+     *   <li>known value, or new value while under quota → returned unchanged (and recorded);</li>
+     *   <li>new value once the quota is reached → returned as {@link #QUOTA_EXCEEDED_VALUE}, and a warning
+     *       is raised once per label.</li>
+     * </ul>
+     */
+    private String maskIfOverQuota(ExecutionContext executionContext, String execId, String metricName,
```
jeromecomte, line 362: It took me time to understand that value means labelValue and not the value of the metric itself

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesMetricSamplesHandler.java
```diff
@@ -146,9 +220,88 @@ private BucketAttributes metricSampleToBucketAttributes(StepMetricSample mm) {
...
+            .computeIfAbsent(labelName, k -> ConcurrentHashMap.newKeySet());
+        if (values.contains(value)) {
+            return value;
+        }
+        // Serialize the size-check-and-add so concurrent live ingestion can't overshoot the quota.
+        synchronized (values) {
+            if (values.contains(value)) {
+                return value;
+            }
+            if (values.size() < maxUniqueLabelValues) {
```
jeromecomte: Here we do the check to ensure that the number of label values per label remains under the defined max. Don't we want to also ensure that we don't exceed the number of labels?

step-controller/step-controller-server/src/main/java/step/core/execution/notices/ExecutionNoticeResolver.java
```diff
@@ -0,0 +1,132 @@
...
+        return new ResolvedExecutionNotice(notice.getTypeId(), UNKNOWN_TYPE_CATEGORY, ExecutionNoticeSeverity.INFO, message, notice.getTimestamp());
+    }
+
+    private String renderMessage(String template, Map<String, String> parameters) {
+        if (template == null) {
+            return "";
+        }
+        Map<String, String> params = (parameters != null) ? parameters : Collections.emptyMap();
+        Matcher matcher = PLACEHOLDER.matcher(template);
+        StringBuffer result = new StringBuffer();
```
jeromecomte: I learnt many years ago that StringBuilder is more efficient. Not sure if still the case in 2026 and if this method is performance sensible

step-controller/step-controller-server/src/main/java/step/core/execution/notices/ResolvedExecutionNotice.java
```diff
@@ -0,0 +1,90 @@
...
+import step.core.execution.model.ExecutionNoticeSeverity;
+
+/**
+ * Read-time, fully resolved view of an execution notice: the message has been rendered from the
+ * notice type's template and the instance parameters (with parameter values HTML-escaped). This is
+ * the shape returned by the REST endpoints; it is never persisted.
+ */
+public class ResolvedExecutionNotice {
+
+    private String typeId;
```
jeromecomte: Could use public fields instead

step-plans/step-plans-core/src/main/java/step/core/execution/model/ExecutionNotice.java
```diff
@@ -0,0 +1,81 @@
...
+ * <p>
+ * A notice only references a registered notice type by its id and carries the parameters used to
+ * resolve the type's parametrized message. The message template, severity and category are held by
+ * the notice type (in the registry) and are resolved server-side at read time. The resolved message
+ * is never persisted, which keeps this model free of any dependency on the registry and allows
+ * message templates to be fixed (or localized) retroactively for already-finished executions.
+ */
+public class ExecutionNotice {
+
+    private String typeId;
```
jeromecomte: Could use public fields instead of getters / setters or maybe records?

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesMetricSamplesHandler.java
david-stephan: refactored

step-controller/step-controller-server/src/main/java/step/core/execution/notices/ExecutionNoticeResolver.java
david-stephan: You're right, this should be a StringBuilder

step-controller-plugins/step-controller-plugins-timeseries/src/main/java/step/plugins/timeseries/TimeSeriesMetricSamplesHandler.java
david-stephan: added now also a quota on the number of labels per exec id and measurement/metric name

step-plans/step-plans-core/src/main/java/step/core/execution/model/ExecutionNotice.java
david-stephan: Actually using public field is an anti pattern and should not be used here. Luckily now that we moved to target Java 21 we can use the recommended approach for immutable Pojo with (de)serialization support: records, I made the change.

step-controller/step-controller-server/src/main/java/step/core/execution/notices/ResolvedExecutionNotice.java
david-stephan: Using public mutable fields breaks encapsulation, making code fragile because any external class can modify the DTO's data at runtime. If the goal was to achieve immutability (the primary reason someone would use a constructor-based @JsonCreator), making the fields public completely defeated that purpose. Now with target Java 21 we can use "record" instead.

jeromecomte (approved):

PR #670, opened 2026-07-02
SED-4781-support-text-preview-for-non-streaming-attachments

cl-exense:


step-controller/step-controller-server/src/main/java/step/resources/ResourceServices.java
```diff
@@ -204,6 +217,72 @@ public Response getResourceContent(@PathParam("id") String resourceId, @QueryPar
...
+    }
+
+    // Implemented as POST because GET would restrict the number of IDs (because of URL length limitations).
+    @POST
+    @Path("/bulk-preview-content")
+    @Secured(right = RESOURCE_RIGHT_NAME + RIGHT_SEPARATOR + READ_RIGHT)
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    public List<ResourcePreview> bulkPreviewContent(List<String> resourceIds, @QueryParam("numberOflines") @DefaultValue("6") int numberOfLines) {
+        if (resourceIds == null) {
```
david-stephan, line 231: Should we raise a bad request if the list is empty?

david-stephan (approved):
Approved with 1 minor comment

step-controller/step-controller-server/src/main/java/step/resources/ResourceServices.java
cl-exense: I don't think that's necessary -- if given an empty list it will simply return an empty list without actually doing any work. Yes, the request will be unnecessary, but there's nothing inherently wrong with it.

PR #672, opened 2026-07-13
SED-4005-added nextExecutionTimestamp on schedulers

lucianbaciu:


step-controller/step-controller-server/src/main/java/step/core/scheduler/CronUtils.java
```diff
@@ -0,0 +1,51 @@
+package step.core.scheduler;
```
david-stephan: This whole class is not required as per my understanding. As discussed, we only want and need to show the next execution time for active schedules and can retrieve the information directly from the quarzt scheduler scheduler.getTriggersOfJob(jobKey) → Trigger.getNextFireTime() 
We should not have to reinvent/ reimplement the code that create divergence. The code itself in this class creates 2 paths for schedule with and without exclusion which is also no required....

step-controller/step-controller-server/src/main/java/step/core/scheduler/SchedulerPlugin.java
```diff
@@ -47,7 +47,8 @@ public void serverStart(GlobalContext context) throws Exception {
         controllerSettingAccessor = context.require(ControllerSettingAccessor.class);
         Collection<ExecutiontTaskParameters> collectionDriver = context.getCollectionFactory().getCollection(EntityConstants.tasks,
             ExecutiontTaskParameters.class);
-        context.get(TableRegistry.class).register(EntityConstants.tasks, new Table<>(collectionDriver, "task-read", true));
+        context.get(TableRegistry.class).register(EntityConstants.tasks, new Table<>(collectionDriver, "task-read", true)
```
david-stephan, line 50: Please follow the same pattern as for ExecutionWrapper and its usage in ExecutionPlugin, we do not want to modify the actual entities.

david-stephan (changes requested):
@lucianbaciu check the 2 main comments requiring changes.

step-controller/step-controller-server/src/main/java/step/core/scheduler/SchedulerPlugin.java
lucianbaciu: @david-stephan please take a look on the updates

step-controller/step-controller-server/src/main/java/step/core/scheduler/ExecutionScheduler.java
```diff
@@ -159,6 +159,14 @@ public String executeExecutionTask(String executionTaskID, String user) {
         return executor.execute(task);
     }
 
+    public Long getNextExecutionDate(String executionTaskID) {
+        ExecutiontTaskParameters task = get(executionTaskID);
+        if (!isSchedulerEnabled() || !task.isActive()) {
```
david-stephan: Minor but the check isSchedulerEnabled could be done before fetching the schedule from the DB. I know it's not done in other method of this class, but we should handle error cases where the task is not found i.e. task is null (e.g. throw a clear exception).

step-controller/step-controller-server/src/main/java/step/core/scheduler/ExecutionScheduler.java
```diff
@@ -159,6 +159,14 @@ public String executeExecutionTask(String executionTaskID, String user) {
         return executor.execute(task);
     }
 
+    public Long getNextExecutionDate(String executionTaskID) {
```
david-stephan: Checking the usages, we should actually pass the task and not its ID, otherwise we fetch it twice from DB for no good reason.

step-controller/step-controller-server/src/main/java/step/core/scheduler/SchedulerPlugin.java
david-stephan: looks correct.

step-controller/step-controller-server/src/main/java/step/core/scheduler/SchedulerServices.java
```diff
@@ -120,6 +120,15 @@ public String executeTask(@PathParam("id") String executionTaskID) {
...
     }
 
+    @Operation(description = "Returns the next execution date of the given scheduler task as a timestamp.")
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    @Path("/{id}/next-execution-date")
+    @Secured(right = "{entity}-read")
+    public Long getNextExecutionDate(@PathParam("id") String executionTaskID) {
+        return scheduler.getNextExecutionDate(executionTaskID);
+    }
```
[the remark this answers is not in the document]
david-stephan: while it's not done for all services, we should do it here, getEntity(executionTaskID); should return a clean exception when the task does not exists + it can be used to directly pass the schedule to getNextExecutionDate (see other comments).

step-controller/step-controller-server/src/main/java/step/core/scheduler/SchedulerTaskTableEnricher.java
```diff
@@ -0,0 +1,40 @@
...
+
+    public SchedulerTaskTableEnricher(Supplier<ExecutionScheduler> executionSchedulerSupplier) {
+        this.executionSchedulerSupplier = executionSchedulerSupplier;
+    }
+
+    @Override
+    public SchedulerTaskWrapper apply(SchedulerTaskWrapper task, Session<?> session, TableParameters tableParameters) {
+        if (task != null) {
+            String cronExpression = task.getCronExpression();
+            if (!task.isActive() || cronExpression == null || cronExpression.isEmpty()) {
```
david-stephan: I'm not sure to understand this list of conditions and don't think any logic to get the next execution time should be implemented in the enricher itself. We already have the checks on task.isActive and the whole logic implemented in scheduler.getNextExecutionDate.

david-stephan (changes requested):

lucianbaciu:
@david-stephan you can recheck

david-stephan (approved):

PR #675, opened 2026-07-17
SED-4846 fix-tenant-header-loss-in-nested-remote-clients

david-stephan:


step-controller/step-controller-remote-client/src/main/java/step/client/AbstractRemoteClient.java
```diff
@@ -64,21 +64,40 @@ public class AbstractRemoteClient implements Closeable {
...
+        // EE only: scope all requests to the configured tenant. In OS context the tenant is null
+        // and no header is sent.
+        String tenant = configuration.getTenant();
+        if (tenant != null && !tenant.isEmpty()) {
+            headers.addProjectName(tenant);
+        }
         if (credentials.getUsername() != null && !credentials.getUsername().trim().isEmpty()) {
             login();
         }
     }
```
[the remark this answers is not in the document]
david-stephan: not new but agreed, updated

step-controller/step-controller-remote-client/src/main/java/step/client/RemoteClientConfiguration.java
```diff
@@ -0,0 +1,65 @@
...
+    private final String tenant;
+
+    public RemoteClientConfiguration(ControllerCredentials credentials) {
+        this(credentials, null);
+    }
+
+    public RemoteClientConfiguration(ControllerCredentials credentials, String tenant) {
+        this.credentials = credentials;
+        this.tenant = tenant;
+    }
```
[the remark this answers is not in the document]
david-stephan: not new but agreed, updated

step-cli/step-cli-core/src/main/java/step/cli/AbstractCliTool.java
```diff
@@ -93,20 +93,19 @@ protected ControllerCredentials getControllerCredentials() {
...
-        RemoteAutomationPackageClientImpl client = new RemoteAutomationPackageClientImpl(getControllerCredentials());
-        addProjectHeaderToRemoteClient(client);
-        return client;
+    /**
+     * @return a {@link RemoteClientConfiguration} carrying both the credentials and the tenant
+     * (EE project name, if configured). Building clients from this configuration ensures the tenant
+     * header is set on the client and inherited by any nested client it creates.
+     */
+    protected RemoteClientConfiguration getRemoteClientConfiguration() {
+        String stepProjectName = parameters.getStepProjectName();
```
jeromecomte, line 102: Shouldn't we call the parameter "tenant" in order to also support non-project tenants like [all]? or do we want to restrict at least in terms of naming to projects? as this API is mainly for internal use, I would probably be in favour of keeping this distinction between tenant and project

step-controller/step-controller-remote-client/src/main/java/step/client/executions/RemoteExecutionManager.java
```diff
@@ -48,7 +51,14 @@
  * This class provides an API for the execution of plans existing on a remote controller
  *
  */
-public class RemoteExecutionManager extends AbstractRemoteClient {
+public class RemoteExecutionManager extends AbstractRemoteClient implements ExecutionProvider {
+
+    /**
+     * Lazily created report tree accessor scoped to this manager's configuration (including tenant).
+     * It shares this manager's lifecycle and is closed together with it, so {@link #getFuture(String)}
```
jeromecomte: These comments with historical character are very useful for PR reviews but useless for the future ("instead of leaking a new client on each call." refers to how it has been once in the past)

step-controller/step-controller-remote-client/src/main/java/step/client/planrunners/RemotePlanRunner.java
```diff
@@ -100,8 +107,9 @@ public PlanRunnerResult runPlanById(Map<String, String> executionParameters, Str
 
         String executionId = executeRequest(() -> b.post(entity, String.class));
 
-        RemoteReportTreeAccessor reportTreeAccessor = new RemoteReportTreeAccessor(credentials);
-        RemoteExecutionProvider executionProvider = new RemoteExecutionProvider(credentials);
+        // Build the nested clients from this runner's configuration so they inherit the tenant.
```
jeromecomte: Do these 2 instances have to be closed? if yes, this doesn't seem to be the case in RemotePlanRunnerResult

step-controller/step-controller-remote-client/src/main/java/step/client/AbstractRemoteClient.java
```diff
@@ -64,21 +65,40 @@ public class AbstractRemoteClient implements Closeable {
 
     protected ControllerCredentials credentials;
 
+    protected RemoteClientConfiguration configuration;
+
+    /**
+     * @deprecated Prefer {@link #AbstractRemoteClient(RemoteClientConfiguration)}, which also
+     * carries the tenant so it (and any nested client) is scoped to the correct project.
+     * This constructor is kept for backward compatibility and creates a tenant-less configuration.
```
jeromecomte, line 73: If we decide to keep it for backward compatibility, we should ensure that the old way of setting a tenant still works and ideally updates the RemoteClientConfiguration. If this only support tenant-less configuration, it would be safer to simply delete it in order to avoid unexpected behaviours with unset tenants

step-cli/step-cli-core/src/main/java/step/cli/AbstractCliTool.java
david-stephan: I would not refactor this legacy code as part of this PR. The parameter 
--projectName=<stepProjectName>
                            The target project name in Step
is used all over the CLI/maven plugin and concretely we have no use case for the [All] tenant in this context.

step-controller/step-controller-remote-client/src/main/java/step/client/executions/RemoteExecutionManager.java
david-stephan: cleaned up the comment

step-controller/step-controller-remote-client/src/main/java/step/client/planrunners/RemotePlanRunner.java
david-stephan: The class is marked as deprecated, so I did not solve the legacy leaks for this one, but I did it now with similar implementation as borh the RemoteExecutionManager

step-controller/step-controller-remote-client/src/main/java/step/client/AbstractRemoteClient.java
david-stephan: The legacy code path hasn't changed whether you're meaning using AdditionalHeader.addProjectName or the RemoteMultitenancyClientImpl.selectTenant and still works the same way as before (I'll add an acceptance criteria for the tester to do some regression tests if required). While our CLI tools now only uses the new mechanism, exiting code (StepClient, customer custom implementation....) still use the legacy paths, we cannot remove it at least not for a minor release. Or did I misunderstood you ?

jeromecomte (approved):

PR #680, opened 2026-07-30
SED-4834 distribution runtime environment and cli integration

cl-exense:


step-cli/step-cli-launcher/src/main/java/step/cli/ApCommand.java
```diff
@@ -169,10 +173,9 @@ protected void executeTool(final String stepUrl, ApDeployParameters params) {
     }
 
     @CommandLine.Command(name = "execute",
-        mixinStandardHelpOptions = true,
-        version = "step.ap.execute 1.0",
         description = "The CLI interface to execute automation packages in Step",
-        usageHelpAutoWidth = true,
+        version = "step.ap.execute 1.0",
```
david-stephan: This isn't new but I would assume it should be aligned to Constants.STEP_VERSION_STRING, no?

step-ide/src/main/resources/logback.xml
```diff
@@ -8,7 +8,10 @@
   <!-- loggers with deviating settings for specific packages -->
   <!--	<logger name="step.parameter.ParameterManager" level="debug" />-->
 
+  <!-- TODO SED-4429: revise these error levels before release -->
   <root level="info">
     <appender-ref ref="STDOUT"/>
   </root>
+  <logger name="step.ide" level="debug"/>
```
david-stephan, line 15: probably not something that should end up in the master

step-cli/step-cli-launcher/src/main/java/step/cli/IdeCommands.java
```diff
@@ -0,0 +1,254 @@
...
+                return CommandLine.ExitCode.SOFTWARE; // (=1)
+            }
+        }
+
+        protected void validateArguments() throws CommandLine.ParameterException {
+            // overridden in subclasses as needed
+        }
+
+        protected void startBackend() throws Exception {
+            LocalIDEState state = LocalIDEState.get();
```
david-stephan: shouldn't we use getState()?

step-cli/step-cli-launcher/src/main/java/step/cli/IdeCommands.java
```diff
@@ -0,0 +1,254 @@
...
+        @Override
+        protected void validateArguments() throws CommandLine.ParameterException {
+            try {
+                var ideState = getState();
+                if (initGroup == null || !initGroup.initialize) {
+                    ideState.validateExistingAutomationPackageDirectory(apDirectory);
+                    return;
+                }
+                // initialization requested
+                Path existingDescriptor = ideState.validateInitializableAutomationPackageDirectory(apDirectory);
```
david-stephan: The method "validateInitializableAutomationPackageDirectory" actually doesn't do the validation or at least not the final one on whether or not a descriptor already exists. The 2 usages looks pretty similar checking and throwing an exception themselves. I see that we have this initGroup.force option that we must handle, but I would rather move the validatio to validateInitializableAutomationPackageDirectoryn, raise a new DescriptorAlreadyExsistExcetion and just catch it here and react depending on the force option

step-cli/step-cli-launcher/src/main/java/step/cli/IdeCommands.java
```diff
@@ -0,0 +1,254 @@
...
+                // Wait for the backend to start, with timeout (see above note)
+                awaitStartup.get(timeoutSeconds, TimeUnit.SECONDS);
+            } catch (TimeoutException e) {
+                throw new Exception("Backend failed to start within the " + timeoutSeconds + " second timeout", e);
+            } catch (ExecutionException e) {
+                throw new Exception("Backend startup failed with an exception", e.getCause());
+            } catch (InterruptedException e) {
+                // Best practice: restore the interrupted status if we catch an InterruptedException
+                Thread.currentThread().interrupt();
+                throw new Exception("Thread was interrupted while waiting for backend to start", e);
```
david-stephan: Exception seem pretty generic, maybe IllegalStateException ? ... details

step-cli/step-cli-launcher/src/main/java/step/cli/IdeCommands.java
```diff
@@ -0,0 +1,254 @@
...
+                logger.info("Your browser should have opened the IDE. If it hasn't, please manually navigate to: {}", url);
+            } catch (Exception e) {
+                logger.warn("Failed to launch browser: {}. Please manually navigate to: {}", e.getMessage(), url);
+            }
+        }
+
+        private static int determineFrontendPort() {
+            // This uses some hardcoded logic, names and port numbers:
+            // If the app is bundled, return prod port, otherwise dev port.
+            String resourceName = "/dist/step-ide/index.html";
```
david-stephan, line 122: not a fan of having such dev mode in code.  I assume it only impact  impact the URL used to open the browser automatically. So if a developer want to use its local frontend, he can just enter the URL directly, right?

step-cli/step-cli-launcher/src/main/java/step/cli/IdeCommands.java
```diff
@@ -0,0 +1,254 @@
...
+            boolean resourceExists = IdeCommands.class.getResource(resourceName) != null;
+            return resourceExists ? 8080 : 4201;
+        }
+
+        @Override
+        public IDEExecutorDelegate createIDEExecutorDelegate(File apFolder, ExecutionParameters executionParams) {
```
david-stephan: I have the feeling folder was refactored to directory, so I guess this one should too ?

step-cli/step-cli-launcher/src/main/java/step/cli/IdeCommands.java
```diff
@@ -0,0 +1,254 @@
...
+                .setWaitForExecution(false)
+                .setEnsureExecutionSuccess(false)
+                .setIncludePlans(executionParams.getDescription()) // NOTE: the include plans is slightly buggy as it splits plan names by commas (",") -- so don't use plan names with a comma.
+                .setExcludePlans(null)
+                .setIncludeCategories(null)
+                .setExcludeCategories(null)
+                .setWrapIntoTestSet(false)
+                .setNumberOfThreads(null)
+                .setReports(null)
+                .setReportOutputDir(Failable.call(() -> Files.createTempDirectory("step-ide-execution-").toFile())); // FIXME clean up tmp dir
```
david-stephan: we don't requests to write any reports (setReports(null)) so report output dir is not required/relevant

step-ide/src/main/java/step/ide/api/LocalIDEServices.java
```diff
@@ -61,19 +70,40 @@ public void initializeNewAP(@QueryParam("existingEmptyDirectory") String existin
...
+        try {
+            // more validation
+            if (LocalIDEState.get().validateInitializableAutomationPackageDirectory(path) != null) {
+                throw new WebApplicationException(
+                    "Directory already contains an automation package descriptor: " + path.toAbsolutePath(),
+                    Response.Status.BAD_REQUEST
+                );
+            }
+
+            if (Files.exists(path) && !Files.isWritable(path)) {
```
david-stephan: Isn't this checked in resolveAndCheckBaseDirectory that is called already by validateInitializableAutomationPackageDirectory?

david-stephan (changes requested):

step-cli/step-cli-launcher/src/main/java/step/cli/ApCommand.java
cl-exense: I have no idea why it was like that ;-) - but yeah, everything else uses the step version, so I'll align it.

step-ide/src/main/resources/logback.xml
cl-exense: Yeah, that's why I kept the SED-4429 comment so we can find it before merging. Anyway, this only affects development, not the distribution.

step-cli/step-cli-launcher/src/main/java/step/cli/IdeCommands.java
cl-exense: It is indeed, but it's directly caught and logged anyway in the caller. Replaced by RuntimeException anyway.

step-cli/step-cli-launcher/src/main/java/step/cli/IdeCommands.java
cl-exense: Yes, it pretty much only affects usage directly from the (dev) IDE, this won't affect any released product. It's still convenient because it opens on the correct URL even while developing ;)

Not sure if I understand your question correctly, but you probably could actually start the released CLI->IDE (which runs on port 8080, including the frontend app), and at the same time a dev frontend on 4201 that will interact with it, and use either of them.

step-cli/step-cli-launcher/src/main/java/step/cli/IdeCommands.java
cl-exense: Hehe, yes, I changed all existing classes except my own :D

step-cli/step-cli-launcher/src/main/java/step/cli/IdeCommands.java
cl-exense: OK, all the better!

step-ide/src/main/java/step/ide/api/LocalIDEServices.java
cl-exense: You're right, that check was present here before and has also been included in the new method. Removing as redundant.

step-cli/step-cli-launcher/src/main/java/step/cli/IdeCommands.java
david-stephan: I mean it is still executed in the productive release even if, **theoretically**, it should not have an impact.
Yes, I meant that it is not really required since the dev can just enter the URL directly in the browser.
But this minor....

david-stephan (approved):

PR #687, opened 2026-08-05
SED-4834 distribution runtime environment and cli integration

cl-exense:


step-ide/src/main/java/step/ide/LocalIDEState.java
```diff
@@ -195,8 +208,19 @@ public String getCurrentAutomationPackageName() {
...
-        return "FIXME";
+        try {
+            Path apDescriptor = Objects.requireNonNull(findAutomationPackageDescriptorPath(currentAutomationPackageDirectory),
+                "Unexpected: unable to find automation package descriptor in " + currentAutomationPackageDirectory);
+            ObjectMapper mapper = AutomationPackageDescriptorReader.createBasicYamlObjectMapper();
+            try (InputStream is = Files.newInputStream(apDescriptor)) {
+                JsonNode rootNode = mapper.readTree(is);
+                JsonNode nameNode = rootNode.get("name");
+                return nameNode != null ? nameNode.asText() : "WARNING: AP name not set";
+            }
```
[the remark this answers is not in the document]
cl-exense: Rejecting, that is exactly the kind of "unexpected" error we're reporting.

step-ide/src/main/java/step/ide/LocalIDE.java
```diff
@@ -27,7 +27,7 @@ public LocalIDE() throws Exception {
         LocalIDEState.get().addDirectoriesToCleanupOnShutdown(List.of(resourcesDirectory, fileManagerDirectory));
         configuration.putProperty("resources.dir", resourcesDirectory.toString());
         configuration.putProperty("grid.filemanager.path", fileManagerDirectory.toString());
-        configuration.putProperty("ui.resource.root", "dist/step-ide");
+        configuration.putProperty("ui.resource.root", LocalIDEState.getIdeResourcePath());
```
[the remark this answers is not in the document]
cl-exense: Fixed.

step-ide/src/main/java/step/ide/LocalIDEState.java
```diff
@@ -195,8 +208,19 @@ public String getCurrentAutomationPackageName() {
         if (currentAutomationPackageDirectory == null) {
             return null;
         }
-        // FIXME: determine name
-        return "FIXME";
+        try {
```
jeromecomte, line 209: Shouldn't we move this code to the AutomationPackageReader instead of implementing the parsing here specifically for the IDE?

jeromecomte (approved):
Approving with a remark relevant for the longterm

step-ide/src/main/java/step/ide/LocalIDEState.java
cl-exense: In principle, I agree. The problem is that instantiating a full reader is complicated, and here, unnecessary. In this context, we explicitly don't care about the full parsing etc. of the package, all we want is to retrieve a single attribute from a single yaml file. What we could do though is instead of (or in addition to) having the static helper method on AutomationPackageDescriptorReader to get a properly configured ObjectMapper, have such static methods for retrieving just the name of an AP (given either as a path/directory, or file, or InputStream). That would still keep the logic simple, but put it closer to where it "belongs".

Makes sense?

jeromecomte: My only concern is to have the code to read AP (which depends on how they are structured) distributed / duplicated in many places. If the read meethod of the reader is too heavy, I would simply move this simplified one to the reader

cl-exense: Agree. Added `getAutomationPackageName` utilty methods to the reader class.

jeromecomte (approved):

PR #688, opened 2026-08-07
SED-4878-use-of-groovy-templates-causes-massive-evaluation-delays

cl-exense:


jeromecomte (changes requested):
Agree with the fix but I would retarget it to the branch master and add a time measurement in the corresponding JUnit to ensure we catch the performance issue in the future

jeromecomte (approved):

PR #690, opened 2026-08-18
SED-3310 Local execution of AP should spawn and execute keywords on actual agents

david-stephan:


step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProcess.java
```diff
@@ -0,0 +1,191 @@
...
+    private static final long FORCIBLE_TERMINATION_TIMEOUT_MS = 5000;
+    private static final int WORKING_DIRECTORY_DELETION_RETRIES = 5;
+    private static final long WORKING_DIRECTORY_DELETION_RETRY_WAIT_MS = 100;
+
+    private final String name;
+    private final Process process;
+    private final Path workingDirectory;
+    private final boolean printOutput;
+    private final Deque<String> lastOutputLines = new ArrayDeque<>();
+    private final CompletableFuture<Void> outputPump;
```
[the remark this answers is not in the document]
david-stephan: While the observation about interruption was not really correct, changes have been made to switch to daemon thread together with other improvements

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProcess.java
```diff
@@ -0,0 +1,191 @@
...
+                        if (lastOutputLines.size() > RETAINED_OUTPUT_LINES) {
+                            lastOutputLines.removeFirst();
+                        }
+                    }
+                }
+            } catch (IOException e) {
+                logger.debug("Error while reading the output of {}", name, e);
+            }
+        });
+    }
```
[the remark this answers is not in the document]
david-stephan: same reply as on previous comment

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProcess.java
```diff
@@ -0,0 +1,191 @@
...
+            }
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            logger.warn("Interrupted while waiting for {} to stop. Destroying it forcibly.", name);
+            destroyForciblyWithDescendants();
+            return;
+        } finally {
+            // The reader terminates on its own once the process closes its output stream, this is only to make sure
+            // the thread is not left running should that not happen.
+            outputPump.cancel(true);
```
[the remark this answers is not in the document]
david-stephan: same reply as on previous comment

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalTokenSelectionCriteriaFilter.java
```diff
@@ -0,0 +1,85 @@
...
+        Map<String, Interest> supported = new LinkedHashMap<>();
+        // Sorted, so that the same set of dropped criteria is reported once whatever the order they were collected in
+        Map<String, String> dropped = new TreeMap<>();
+        selectionCriteria.forEach((criterion, interest) -> {
+            if (SUPPORTED_CRITERIA.contains(criterion)) {
+                supported.put(criterion, interest);
+            } else {
+                dropped.put(criterion, interest.getSelectionPattern().pattern());
+            }
+        });
```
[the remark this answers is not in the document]
david-stephan: updated

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalProcessAgentProvisioningDriver.java
```diff
@@ -0,0 +1,446 @@
...
+            // Sequentially: there are at most as many agents as there are agent types, and starting them one after
+            // the other keeps the output of a failing one readable.
+            for (Map.Entry<String, Integer> entry : tokensByAgentType.entrySet()) {
+                String agentType = entry.getKey();
+                int numberOfTokens = entry.getValue();
+                AgentPoolProvisioningReport poolReport = new AgentPoolProvisioningReport(
+                    new AgentPoolRequirementSpec(agentPoolName(agentType), 1));
+                report.addAgentPoolReport(poolReport);
+                try {
+                    session.startedAgents.add(startAgent(agentType, numberOfTokens, session.tokenPartition));
```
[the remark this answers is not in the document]
david-stephan: The whole call stack is single threaded and ordered, not applicable.

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalProcessAgentProvisioningDriver.java
```diff
@@ -0,0 +1,446 @@
...
+        ProvisioningSession session = provisioningSessions.get(provisioningRequestId);
+        return session != null ? session.status : null;
+    }
+
+    @Override
+    public void deprovisionTokens(String provisioningRequestId) {
+        ProvisioningSession session = provisioningSessions.remove(provisioningRequestId);
+        if (session != null) {
+            stopAgents(session);
+        }
```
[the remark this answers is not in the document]
david-stephan: same reply as on previous comment

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalProcessAgentProvisioningDriver.java
```diff
@@ -0,0 +1,446 @@
...
+    }
+
+    /**
+     * Stops whatever is still running. Normally a no-op, as the agents of an execution are stopped when it is
+     * de-provisioned; it covers the executions which never got that far.
+     */
+    @Override
+    public void close() {
+        provisioningSessions.values().forEach(this::stopAgents);
+        provisioningSessions.clear();
```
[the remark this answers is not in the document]
david-stephan: same reply as on previous comment

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/ScriptEngineLibraries.java
```diff
@@ -0,0 +1,323 @@
...
+            temporaryDirectory = Files.createTempDirectory(directory.getParent(), directory.getFileName() + ".part");
+        } catch (IOException e) {
+            throw new LocalAgentException("Error while creating the directory of the " + engine.language() + " libraries", e);
+        }
+        try {
+            logger.info("Extracting the {} libraries to {}...", engine.language(), directory);
+            for (URL library : libraries.values()) {
+                copyLibrary(library, temporaryDirectory);
+            }
+            Files.move(temporaryDirectory, directory, StandardCopyOption.ATOMIC_MOVE);
```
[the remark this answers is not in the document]
david-stephan: updated

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/DotNetLocalAgentProvider.java
```diff
@@ -0,0 +1,235 @@
...
+     * provider starts the agent directly. A failure here is not reported as an error of its own, the process start
+     * failing right after with the message of the operating system.
+     */
+    private static void makeExecutable(Path file) {
+        if (WINDOWS || Files.isExecutable(file)) {
+            return;
+        }
+        logger.debug("Making {} executable", file);
+        if (!file.toFile().setExecutable(true)) {
+            logger.warn("Unable to make {} executable. Run 'chmod +x' on it if the agent fails to start.", file);
```
cl-exense: You'll want to log file.toAbsolutePath() just to be sure.

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/JavaLocalAgentProvider.java
```diff
@@ -0,0 +1,263 @@
...
+     * Extracts the embedded agent bundle, once per CLI version. The extracted bundle is kept between runs: it is
+     * ~20 MB and rewriting it on every local execution would be pure waste.
+     */
+    private Path extractEmbeddedAgent() throws LocalAgentException {
+        Path installedDirectory = workspace.getInstalledAgentDirectory(INSTALLED_AGENT_NAME, Constants.STEP_VERSION_STRING);
+        Path agentJar = installedDirectory.resolve(INSTALLED_AGENT_JAR_NAME);
+        // The directory is named after the version, which does not change between two builds of the same one. An
+        // extracted agent is therefore only reused when it is also the same size as the embedded one, otherwise
+        // every CLI rebuilt during development would go on running the agent of the first build, for ever.
+        long embeddedSize = embeddedAgentSize();
```
cl-exense: This is not guaranteed to never produce "false negatives". While it's unlikely that the size remains the same over different builds, it's not impossible. I suggest to check the size AND the checksum to minimize the risk; the overhead will be extremely low (compared to everything else we're doing). Here's some simple code:

```
public long checksum(InputStream is) throws Exception {
    try (CheckedInputStream checkedStream = new CheckedInputStream(is, new CRC32C())) {
        checkedStream.transferTo(OutputStream.nullOutputStream());
        return checkedStream.getChecksum().getValue();
    }
}
```

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/JavaLocalAgentProvider.java
```diff
@@ -0,0 +1,263 @@
...
+            logger.debug("Unable to determine the size of the embedded Java agent", e);
+            return -1;
+        }
+    }
+
+    private static long sizeOf(Path file) {
+        try {
+            return Files.size(file);
+        } catch (IOException e) {
+            logger.debug("Unable to determine the size of {}", file, e);
```
cl-exense: Use .toAbsolutePath() to be sure.

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/JavaLocalAgentProvider.java
```diff
@@ -0,0 +1,263 @@
...
+     */
+    private static String currentJavaExecutable() throws LocalAgentException {
+        return ProcessHandle.current().info().command().orElseThrow(() -> new LocalAgentException(
+            "Unable to determine the java executable of the current process, which is needed to start the local Java agent."));
+    }
+
+    private List<String> vmArgs() {
+        return Optional.ofNullable(configuration.getJavaAgentVmArgs())
+            .filter(args -> !args.isBlank())
+            .map(args -> List.of(args.trim().split("\\s+")))
```
cl-exense: This is reasonable, but still dangerous. Are we sure that individual arguments inside the `args` String can never contain spaces? (Contrived example for a shell: `java -cp "C:/Program Files/crash boom bang/*"`

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentPoolRequirements.java
```diff
@@ -0,0 +1,108 @@
...
+     */
+    static List<AgentPoolRequirementSpec> forRequiredAgentTypes(Collection<Function> functions,
+                                                               FunctionTypeRegistry functionTypeRegistry,
+                                                               Set<String> availableAgentTypes, int tokensPerAgent) {
+        Set<String> requiredAgentTypes = requiredAgentTypes(functions, functionTypeRegistry);
+        requiredAgentTypes.retainAll(availableAgentTypes);
+        if (requiredAgentTypes.isEmpty()) {
+            // Either the keywords are unknown at this point, or none of them declares an agent type this machine can
+            // serve. Starting everything available is the only answer left which lets the execution run at all.
+            logger.debug("No agent type could be derived from the keywords. Starting one agent of each available type.");
```
cl-exense: Is that a meaningful tactic, i.e. does it actually solve the problem? IOW: is "we don't know the keywords that will run" a meaningful and expected situation?

I would expect to
1. clearly know what **wants** to run (e.g. ["java", ".net"]). It that's empty, then nothing needs to be spawned.
2. clearly know what from that list **can** be run (i.e. if no .net is installed, then it's simply impossible). If that's empty, there's still no need to spawn anything, because it won't help solve the problem anyway.

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentPoolRequirements.java
```diff
@@ -0,0 +1,108 @@
...
+                tokensPerAgent))
+            .collect(Collectors.toList());
+    }
+
+    /**
+     * @return the agent types the given keywords need, as their function types declare them. Keywords running in the
+     * engine itself, composite keywords being the usual case, need no agent and are left out.
+     */
+    private static Set<String> requiredAgentTypes(Collection<Function> functions, FunctionTypeRegistry functionTypeRegistry) {
+        Set<String> agentTypes = new LinkedHashSet<>();
```
cl-exense: Why a LinkedHashSet specifically? Is order relevant?

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentPoolRequirements.java
```diff
@@ -0,0 +1,108 @@
...
+        for (Function function : functions) {
+            if (function.requiresLocalExecution()) {
+                continue;
+            }
+            try {
+                AbstractFunctionType<Function> functionType = functionTypeRegistry.getFunctionTypeByFunction(function);
+                Map<String, Interest> criteria = functionType.getTokenSelectionCriteria(function);
+                Interest agentType = criteria != null ? criteria.get(AgentTypes.AGENT_TYPE_KEY) : null;
+                if (agentType != null) {
+                    agentTypes.add(agentType.getSelectionPattern().pattern());
```
cl-exense: .pattern() sounds a lot like there could be placeholders inside. If there can, how and where are they resolved / handled?

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProcess.java
```diff
@@ -0,0 +1,226 @@
...
+        // never be what keeps the CLI from exiting. Nothing can stop it from the outside - a blocking read on a
+        // process pipe answers neither to Thread.interrupt() nor to closing the stream, which blocks the caller
+        // instead on Windows - so the end of the process' output is the only thing that ends it.
+        pump.setDaemon(true);
+        pump.start();
+        return pump;
+    }
+
+    private void pumpOutput() {
+        try (BufferedReader reader = new BufferedReader(new InputStreamReader(process.getInputStream(), StandardCharsets.UTF_8))) {
```
cl-exense, line 92: Is forcing UTF-8 actually the correct approach here? I think there are still some weird Windows installations that in fact use some medieval charset by default, and by consequence may produce output in a local charset.

Then again, treating everything as standard UTF-8 and accepting the potential risk of a few garbled characters is also OK with me.

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProcess.java
```diff
@@ -0,0 +1,226 @@
...
+            String line;
+            while ((line = reader.readLine()) != null) {
+                if (printOutput) {
+                    logger.info("{} -- {}", name, line);
+                } else {
+                    logger.debug("{} -- {}", name, line);
+                }
+                synchronized (lastOutputLines) {
```
cl-exense, line 100: use a ConcurrentLinkedDeque instead, and get rid of the synchronized block.

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProcess.java
```diff
@@ -0,0 +1,226 @@
...
+                    lastOutputLines.addLast(line);
+                    if (lastOutputLines.size() > RETAINED_OUTPUT_LINES) {
```
cl-exense, line 102: Another optimization, as .size() is O(n) (yes, we're only talking 50, but there's a simpler and more efficient way):

```
// retainedLinesCount is a field
lastOutputLines.addLast(line);
if (retainedLinesCount < RETAINED_OUTPUT_LINES) {
    retainedLinesCount++;
} else {
    lastOutputLines.removeFirst();
}
```

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProcess.java
```diff
@@ -0,0 +1,226 @@
...
+
+    public boolean isAlive() {
+        return process.isAlive();
+    }
+
+    /**
+     * @return the last lines the agent printed, oldest first. Used to explain a start failure.
+     */
+    public List<String> getLastOutputLines() {
+        synchronized (lastOutputLines) {
```
cl-exense, line 129: Can also be removed if using appropriate thread-safe class.

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProcess.java
```diff
@@ -0,0 +1,226 @@
...
+     * only waits for that shutdown to complete and falls back to killing the process, so that a hanging or already
+     * unreachable agent can never keep the CLI from terminating.
+     *
+     * @param gracefulTimeoutMs how long to wait for the process to terminate on its own. Pass {@code 0} when the
+     *                          agent was not asked to shut down, or refused to: there is then nothing to wait for,
+     *                          and waiting would only delay the end of the execution by the whole timeout.
+     */
+    public void stop(long gracefulTimeoutMs) {
+        try {
+            if (gracefulTimeoutMs > 0 && process.waitFor(gracefulTimeoutMs, TimeUnit.MILLISECONDS)) {
```
cl-exense, line 147: Given the explanations below (in the "killing with descendants" path) concerning the child processes: Are all processes guaranteed to properly terminate everything when this "happy path" works? I.e. waitFor returns true -> everything is cleaned up, no zombie processes? IOW: We trust the processes to perform proper cleanup on normal termination, but are more aggressive and kill the process tree when they fail to terminate. Makes sense I guess, but just wanted to double-check.

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProcess.java
```diff
@@ -0,0 +1,226 @@
...
+     * <p>
+     * The agent is not always the process started here: a globally installed Node.js agent is started through the
+     * command npm put on the PATH, which on Windows is a command file and therefore runs the agent as a child of
+     * cmd.exe. Killing that shell alone would leave the agent behind, holding its port, its registration on the grid
+     * and the files of the working directory this method's caller then tries to delete.
+     */
+    private void destroyForciblyWithDescendants() {
+        // Taken before anything is killed: the descendants of a dead process are no longer reachable through it
+        List<ProcessHandle> descendants = process.descendants().toList();
+        descendants.forEach(ProcessHandle::destroyForcibly);
```
cl-exense: Any chances that destroyFocibly is throwing exceptions? The API doesn't declare any, but given that we're already in an abnormal situation I wouldn't count on it, and feel better if this was wrapped with some exception handling.

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProvisioningConfiguration.java
```diff
@@ -0,0 +1,200 @@
...
+    public LocalAgentProvisioningConfiguration setAgentShutdownTimeout(Duration agentShutdownTimeout) {
+        this.agentShutdownTimeout = agentShutdownTimeout;
+        return this;
+    }
+
+    /**
+     * @return extra JVM arguments for the Java agent, mainly to attach a debugger or tune the heap while
+     * troubleshooting a keyword locally
+     */
+    public String getJavaAgentVmArgs() {
```
cl-exense: If this is the only purpose, and its only use is by the Java agent provider, then I suggest to actually make it a list of strings directly. This entirely avoids the "splitting by spaces" problem I mentioned somewhere above. Yes, it will be minimally less straightforward to use, but it will completely avoid that class of potential bugs while still remaining understandable.

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProvisioningPlugin.java
```diff
@@ -0,0 +1,329 @@
...
+        this.configuration = Objects.requireNonNull(configuration, "configuration must not be null");
+    }
+
+    @Override
+    public void initializeExecutionEngineContext(AbstractExecutionEngineContext parentContext, ExecutionEngineContext context) {
+        if (closed) {
+            // Would otherwise publish the grid client of a grid which has been stopped, and fail every keyword later
+            throw new PluginCriticalException("This local agent provisioning plugin has been closed and cannot be reused");
+        }
+        if (driver != null) {
```
cl-exense: Add a comment along the lines of: if driver is already set up, there is nothing to do. Otherwise do some lazy initialization.

cl-exense: Also note that there is a (theoretical) race condition if this method is called twice in parallel before driver is initialized (both threads would race to initialize stuff), but I don't know if it's realistic or purely hypothetical.

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProvisioningPlugin.java
```diff
@@ -0,0 +1,329 @@
...
+            requiredAgentPools = tokenForecastingContext.getAgentPoolRequirementSpec();
+            Set<Map<String, Interest>> criteriaWithoutMatch = tokenForecastingContext.getCriteriaWithoutMatch();
+            if (!criteriaWithoutMatch.isEmpty()) {
+                // Typically a keyword of a language whose agent this machine has no installation of
+                throw new ProvisioningException(unavailableAgentsMessage(criteriaWithoutMatch, driver::getInstallationHint));
+            }
+        } else {
+            List<AgentPoolRequirementSpec> configuredAgentPools = planAgentConfiguration.getAgentPoolRequirementSpecs();
+            if (configuredAgentPools == null) {
+                throw new ProvisioningException("The agent configuration of the plan returned no agent pool requirement");
```
cl-exense: Rephrase to something like "has automatic calculation disabled but no requirements defined"

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProvisioningPlugin.java
```diff
@@ -0,0 +1,329 @@
...
+        }
+    }
+
+    /**
+     * Turns the criteria no local agent pool matched into an error naming the agent types and, when their provider has
+     * one, what to do about them: the criteria as they are collected ({@code [{$agenttype=dotnet}]}) do say what is
+     * missing, but not that the .NET agent is the user's to install, nor how this CLI is told where it is.
+     */
+    // Package private for the sake of the tests, which cover the message without running an execution
+    // Qualified: step.functions.Function, the keyword, is the Function of this package
```
cl-exense: This comment is more confusing than it needs to be. Either remove entirely, or rename to something like
`// using fully-qualified java...Function because step.functions.Function is already imported`

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProvisioningPlugin.java
```diff
@@ -0,0 +1,329 @@
...
+            .distinct()
+            .collect(Collectors.toList());
+        if (agentTypes.isEmpty()) {
+            // Criteria this plugin cannot read as an agent type, reported as they were collected
+            return "This plan requires agents which are not available for local execution: " + criteriaWithoutMatch;
+        }
+        StringBuilder message = new StringBuilder("This plan requires agent types which are not available for local"
+            + " execution: " + String.join(", ", agentTypes) + ".");
+        agentTypes.stream().map(installationHints).filter(Objects::nonNull)
+            .forEach(hint -> message.append(" ").append(hint));
```
cl-exense: Minor: Instead of concatenating separated by spaces only, you could use
```
... .filter(Objects::nonNull).collect(Collectors.joining("; ", " ", ""))
```

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProvisioningPlugin.java
```diff
@@ -0,0 +1,329 @@
...
+     * their handlers (the composite handler of every plan calling a composite keyword, for one) with class loaders
+     * reading the jars straight out of the file manager directory of this grid. Deleting it before the grid client
+     * released them leaves the files open, and Windows refuses to delete an open file.
+     * <p>
+     * The caller closes this plugin after the execution engine it was given to, typically by declaring it as the first
+     * resource of the same try-with-resources.
+     */
+    @Override
+    public void close() throws IOException {
+        closed = true;
```
cl-exense, line 327: Should this be quarded by a `if (!closed)` ? Edit: mehr, probably not needed.

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProvisioningPlugin.java
```diff
@@ -0,0 +1,329 @@
...
+        if (grid != null) {
+            grid.close();
+            grid = null;
+        }
+    }
+
+    private static void closeQuietly(LocalExecutionGrid grid) {
```
cl-exense, line 334: The method name doesn't match the error message (and the behavior, it's not always quiet). Suggest to rename to `closeFailed` or something along these lines.

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentStartContext.java
```diff
@@ -0,0 +1,76 @@
...
+     * @param gridUrl          the URL of the embedded grid the agent has to register to. Always a loopback URL.
+     * @param workingDirectory the directory the agent process runs in. It is created by the caller, is dedicated to
+     *                         this agent and is deleted once the agent is stopped.
+     * @param numberOfTokens   the capacity of the single token group the agent has to declare
+     * @param tokenAttributes  the attributes to declare on the tokens. Contains at least the token partition, which
+     *                         isolates the tokens of this execution from the ones of any other.
+     * @param gridSecurity     the security configuration of the embedded grid, or {@code null} when the grid
+     *                         authentication is disabled
+     */
+    public LocalAgentStartContext(String gridUrl, Path workingDirectory, int numberOfTokens,
```
cl-exense, line 48: I propose to consistently use records for such data-only classes, as they avoid almost all of the boilerplate and allow for really easy and compact classes:

```
/**
 * Everything a {@link LocalAgentProvider} needs to start one local agent process.
 *
 * @param gridUrl          the URL of the embedded grid the agent has to register to. Always a loopback URL.
 * @param workingDirectory the directory the agent process runs in. It is created by the caller, is dedicated to
 *                         this agent and is deleted once the agent is stopped.
 * @param numberOfTokens   the capacity of the single token group the agent has to declare
 * @param tokenAttributes  the attributes to declare on the tokens. Contains at least the token partition, which
 *                         isolates the tokens of this execution from the ones of any other.
 * @param gridSecurity     the security configuration of the embedded grid, or {@code null} when the grid
 *                         authentication is disabled
 */
public record LocalAgentStartContext(
        String gridUrl, 
        Path workingDirectory, 
        int numberOfTokens,
        Map<String, String> tokenAttributes, 
        SymmetricSecurityConfiguration gridSecurity
) {

    public LocalAgentStartContext {
        Objects.requireNonNull(gridUrl, "gridUrl must not be null");
        Objects.requireNonNull(workingDirectory, "workingDirectory must not be null");
        Objects.requireNonNull(tokenAttributes, "tokenAttributes must not be null");
    }
    
}
```

Not only does that save ~ 30% of space, it also automatically adds useful implementations of `equals`, `hashCode`, and most importantly `toString`.

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentWorkspace.java
```diff
@@ -0,0 +1,131 @@
...
+    private final Path root;
+
+    public LocalAgentWorkspace(Path configuredRoot) throws IOException {
+        this.root = configuredRoot != null ? configuredRoot.toAbsolutePath() : defaultRoot();
+        Files.createDirectories(root);
+        logger.debug("Using local agent workspace {}", root);
+        sweepStaleRunDirectories();
+    }
+
+    private static Path defaultRoot() {
```
cl-exense, line 70: Add a comment (here, or in the class doc) that the default uses a directory inside the dedicated "temporary storage", but always the same. (This is different behavior than asking for a temporary directory, which will always return a new one and would make the "leftovers cleanup" moot).

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentWorkspace.java
```diff
@@ -0,0 +1,131 @@
...
+        return Path.of(System.getProperty("java.io.tmpdir")).resolve(WORKSPACE_DIRECTORY_NAME).toAbsolutePath();
+    }
+
+    /**
+     * Removes the throw-away directories left behind by a previous run. Any directory found here at this point
+     * belongs to a run which is over: the CLI is a short lived process which cleans its own directories up as it
+     * stops the agents and the grid they belong to.
+     */
+    private void sweepStaleRunDirectories() {
```
cl-exense, line 79: Here's the modernized and slightly more efficient version (stream vs File[]):

```
private void sweepStaleRunDirectories() {
    try (DirectoryStream<Path> stream = Files.newDirectoryStream(root, path -> {
        String name = path.getFileName().toString();
        return Files.isDirectory(path) &&
            (name.startsWith(AGENT_DIRECTORY_PREFIX) || name.startsWith(GRID_DIRECTORY_PREFIX));
    })) {

        for (Path staleDirectory : stream) {
            logger.info("Removing the directory {} left over by a previous run.", staleDirectory);
            try {
                FileUtils.deleteDirectory(staleDirectory.toFile());
            } catch (IOException e) {
                logger.warn("Failed to delete the stale directory {}.", staleDirectory, e);
            }
        }

    } catch (IOException e) {
        logger.warn("Failed to read the contents of the root directory {}.", root, e);    
    }
}

```

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalExecutionGrid.java
```diff
@@ -0,0 +1,147 @@
...
+    private final Path fileManagerDirectory;
+
+    /**
+     * @param workspace the workspace the file manager of this grid caches its files in. Using it rather than a
+     *                  temporary directory of its own is what gets that cache deleted: it is deleted with this grid,
+     *                  and swept by the next run should this CLI be killed before it can do so.
+     */
+    public LocalExecutionGrid(Duration agentStartTimeout, LocalAgentWorkspace workspace) throws Exception {
+        // The grid listens on all interfaces, so it is protected with a secret rather than left open to anything
+        // running on the machine. The secret is generated per invocation and never leaves this process and the
```
cl-exense: If it listens on all interfaces, then it is not only open to "anything running on the machine", but to "anything able to connect" (from anywhere), which is a much broader threat theoretically. The comment should be changed accordingly.

I would still prefer to limit the listening interfaces if possible (ideally 127.0.0.1 or ::1), because what's not accessible cannot be exploited, but I guess using security keys is the next best option.

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalExecutionGrid.java
```diff
@@ -0,0 +1,147 @@
...
+     */
+    public String getGridUrl() {
+        return "http://" + AgentConfWriter.LOOPBACK_HOST + ":" + grid.getServerPort();
+    }
+
+    /**
+     * Stops the grid and deletes the files its file manager cached. The grid client is left alone: it is registered
+     * in the execution engine context, which closes it itself.
+     * <p>
+     * To be called once the grid client has been closed, and not before: the local tokens of that client execute
```
cl-exense: Confusing comment phrasing, suggest to rephrase to simply: this method must not be called before the client has been closed, as the client depends on files that this method deletes.

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalPorts.java
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * @return a port which was free on the loopback interface. The socket is closed before returning, so this is a
+     * hint rather than a reservation: the agent binds it moments later, and nothing else on this machine is expected
+     * to take it in between.
+     */
+    static int findFreeLoopbackPort() throws LocalAgentException {
+        try (ServerSocket socket = new ServerSocket(0, 1, InetAddress.getByName(AgentConfWriter.LOOPBACK_HOST))) {
+            return socket.getLocalPort();
+        } catch (IOException e) {
+            throw new LocalAgentException("Unable to reserve a port for a local agent", e);
```
cl-exense: "Unable to find a free port".
After all "this is a hint rather than a reservation" :-P

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProvisioningConfiguration.java
```diff
@@ -0,0 +1,200 @@
...
+    public LocalAgentProvisioningConfiguration setAgentStartTimeout(Duration agentStartTimeout) {
+        this.agentStartTimeout = agentStartTimeout;
+        return this;
+    }
+
+    public Duration getAgentShutdownTimeout() {
+        return agentShutdownTimeout;
+    }
+
+    public LocalAgentProvisioningConfiguration setAgentShutdownTimeout(Duration agentShutdownTimeout) {
```
cl-exense, line 151: Because of how the stop method where this is used is implemented, it will treat 0 values differently from normal timeouts (instead directly runing an unconditional force-kill). Therefore, I think accepting 0 here is not wise -- users might accidentally set that, maybe even thinking it means "unlimited". I suggest to enforce that the argument is at least 1 millisecond.

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalProcessAgentProvisioningDriver.java
```diff
@@ -0,0 +1,446 @@
...
+        final AgentProvisioningRequest request;
+        final String provisioningId = UUID.randomUUID().toString();
+        final AgentProvisioningStatus status = new AgentProvisioningStatus();
+        final Set<StartedAgent> startedAgents = new HashSet<>();
+        /**
+         * Isolates the tokens of this execution from the ones of any other, exactly as the Kubernetes driver does.
+         * The execution id is used when available so that the agent configurations left behind by a crash can be
+         * traced back to the execution they belonged to.
+         */
+        final String tokenPartition;
```
cl-exense, line 430: I don't really understand. Does that mean the partition ID is used e.g. in directory names, and logged (or printed) to the user?

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalTokenSelectionCriteriaFilter.java
```diff
@@ -0,0 +1,95 @@
...
+ */
+public class LocalTokenSelectionCriteriaFilter implements TokenSelectionCriteriaFilter {
+
+    private static final Logger logger = LoggerFactory.getLogger(LocalTokenSelectionCriteriaFilter.class);
+
+    /**
+     * The criteria a local execution can honour: the type of agent the keyword needs, and the partition isolating the
+     * tokens of this execution.
+     */
+    private static final Set<String> SUPPORTED_CRITERIA = Set.of(AgentTypes.AGENT_TYPE_KEY, TOKEN_ATTRIBUTE_PARTITION);
```
cl-exense, line 57: I guess I still don't understand the partition concept. If every execution starts up and provisions its own agents and tears them down afterwards, shouldn't we always have exactly one partition?

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/NodeLocalAgentProvider.java
```diff
@@ -0,0 +1,415 @@
...
+    private static final String INSTALLED_AGENT_NAME = "node";
+    private static final String NODE_MODULES_DIRECTORY_NAME = "node_modules";
+    private static final String BIN_DIRECTORY_NAME = ".bin";
+    private static final String PACKAGE_JSON_FILE_NAME = "package.json";
+    /**
+     * Written before installing, so that npm installs into the directory of this CLI rather than into whichever
+     * project happens to sit above it: npm looks for the nearest {@code package.json} up the directory tree, and a
+     * workspace placed inside a JavaScript project would otherwise be installed into that project.
+     */
+    private static final String PACKAGE_JSON_CONTENT =
```
cl-exense: Minor (no need to fix here, but as a hint for the future): Such kinds of strings can now easily be embedded using the
```
String x = """
multiline
text
""";
```
syntax, which is much more readable.

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/AgentConfWriter.java
```diff
@@ -0,0 +1,103 @@
...
+ */
+public class AgentConfWriter {
+
+    /**
+     * The agents are only ever reached from this machine, over the loopback interface. The host is pinned rather than
+     * left to the agent to determine, because agents fall back to resolving the canonical host name of the machine
+     * (see {@code BaseServer.getOrBuildActualUrl}), which on a VPN, a multi-homed machine or one with a restricted DNS
+     * resolves to a name the grid then fails to call back on.
+     */
+    public static final String LOOPBACK_HOST = "127.0.0.1";
```
cl-exense, line 62: 127.0.0.1 is not guaranteed to be available on all machines -- on an IPv6-only machine the loopback interface is `::1` . An environment-agnostic version is to explicitly use `InetAddress.getLoopbackAddress()`, but I suppose that supporting IPv6-only setups (which are slowly beginning to exist) will require more extensive adaptations and tests in other places, so maybe just add a TODO or FIXME here as a reminder.

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/ScriptEngineLibraries.java
```diff
@@ -0,0 +1,327 @@
...
+        } catch (URISyntaxException e) {
+            throw new LocalAgentException("Unable to read the library " + location, e);
+        }
+    }
+
+    /**
+     * Reads a library out of the archive holding it, given a location of the form
+     * {@code jar:file:/path/app.jar!/BOOT-INF/lib/library.jar!/}.
+     */
+    private static void copyNestedLibrary(URL location, Path target) throws IOException, LocalAgentException {
```
cl-exense, line 279: This method is **assuming** that the location starts with `"jar:"`, but that's never verified. Either the caller, or the method itself, should check the location for that prefix (and probably throw some kind of runtime exception if it isn't). I know it's unlikely, but better safe than sorry.

step-cli/step-cli-launcher/src/main/java/step/cli/StepConsole.java
```diff
@@ -49,6 +57,23 @@ public class StepConsole implements Callable<Integer> {
     public static final Logger log = LoggerFactory.getLogger(StepConsole.class);
     public static final String MANAGED = "managed:";
 
+    /**
+     * The conversion words of logback's exception converters, {@code %ex} and {@code %throwable} with their variants
+     */
+    private static final List<String> EXCEPTION_CONVERSION_WORDS = List.of("%ex", "%exception", "%throwable",
```
cl-exense: The entire logger handling looks a bit like voodoo. I'm fine with keeping it, but suggest to move it into a separate utility class focused on that functionality, if possible.

step-cli/step-cli-launcher/src/test/java/step/cli/StepConsoleTest.java
```diff
@@ -561,6 +561,7 @@ public static class LocalExecutionParams {
             private String includeCategories;
             private String excludeCategories;
             private Map<String, String> executionParameters;
+            public step.agents.provisioning.local.LocalAgentProvisioningConfiguration localAgentConfiguration;
```
cl-exense: needlessly fully-qualified, I suppose?

step-plans/step-plans-core/src/main/java/step/core/execution/OperationMode.java
```diff
@@ -24,15 +24,39 @@ public enum OperationMode {
```
cl-exense, line 30: General note on this class - both the constants and method names are rather confusing, even with the comments. It all sounds like it defines exceptions on top of exceptions on top of exceptions :-)

I can't really think of much better names off the top of my head, but we can have a short discussion/brainstorming if needed.

step-plans/step-plans-base-artefacts/src/main/java/step/engine/plugins/FunctionPlugin.java
```diff
@@ -52,10 +54,12 @@ public class FunctionPlugin extends AbstractExecutionEnginePlugin {
...
         FileResolver fileResolver = context.getFileResolver();
 
-        gridClient = context.inheritFromParentOrComputeIfAbsent(parentContext, GridClient.class, k -> new MockedGridClientImpl());
-        if (parentContext != null) {
-            grid = parentContext.get(Grid.class);
-        }
+        // A plugin initialized before this one may have provided a grid of its own: the CLI does, for the agents it
+        // starts on the developer machine. Since inheritFromParentOrComputeIfAbsent looks at the parent context
+        // alone, such a client is only kept by falling back to it instead of to the mocked one.
+        gridClient = context.inheritFromParentOrComputeIfAbsent(parentContext, GridClient.class,
```
cl-exense, line 59: I don't understand neither the code nor the comment. The code is probably correct, but the comment would be much easier if it explained it as (if I understood it correctly): prefer from parent context, then context, otherwise mocked.

cl-exense (changes requested):
Approving with a few comments. In general, that's very solid and well-documented code, so most comments are more about understandability etc. than actual bugs (IIRC), and only very few of them are actual requests to change something.

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/DotNetLocalAgentProvider.java
david-stephan: done

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/JavaLocalAgentProvider.java
david-stephan: done

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/AgentConfWriter.java
david-stephan: added as TODO

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/JavaLocalAgentProvider.java
david-stephan: refactored to use the checksum available in the JAR (embedded) and storing the current one (extracted) on disk. Note that this only required for local dev, so we should not over engineer and limit the impact on the production code

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/JavaLocalAgentProvider.java
david-stephan: changed to a support a list of VM arg options to avoid the list of splits all together

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProcess.java
david-stephan: yes, I don't think a more complicated implementation has much added value here (that's quite a edge case and the impact is negligeable.

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentPoolRequirements.java
david-stephan: I didn't get that one, but pattern is a regexp pattern, this is legacy step.grid.tokenpool.Interest code

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProvisioningConfiguration.java
david-stephan: that's kind of what I did now, see previous related comment

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentPoolRequirements.java
david-stephan: This path was incorrect, updated

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProcess.java
david-stephan: That's our standard implementation of agent graceful shutdown, the processes are responsible of stopping their own children. We could imagine to get the tree before the graceful shutdown and kill whatever is left afterward, but would not change the behavior here.

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProcess.java
david-stephan: honestly no idea, but adding a catch throwable now

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentPoolRequirements.java
david-stephan: Was purely cosmetics  to list them in messages and start them in the order they are used int the plan, Switched.

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProcess.java
david-stephan: Using ArrayDeque is more efficient in term of memory allocation and comparable in terms of locking, it's size method is actually O(1), keeping current implementation.

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProcess.java
david-stephan: see previous reply

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProcess.java
david-stephan: see previous replies

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProvisioningPlugin.java
david-stephan: done

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProvisioningPlugin.java
david-stephan: removed completely

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProvisioningPlugin.java
david-stephan: skipped

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProvisioningPlugin.java
david-stephan: It's actually the standard for a method attempting to close a resource catching and not rethrowing the exception

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProvisioningPlugin.java
david-stephan: indeed, not needed, closed is a guard for another method

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentStartContext.java
david-stephan: skipped

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentWorkspace.java
david-stephan: extended javadoc

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentWorkspace.java
david-stephan: changed since it has a better error handling than listfiles

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalExecutionGrid.java
david-stephan: comment updated

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalExecutionGrid.java
david-stephan: comment updated

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalPorts.java
david-stephan: Comment updated

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalProcessAgentProvisioningDriver.java
david-stephan: Rewrote the comment

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProvisioningConfiguration.java
david-stephan: changed to a minium of 1 second which is the unit used.

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalAgentProvisioningPlugin.java
david-stephan: The implementation of lazy loading is actually incorrect here. I fixed it. The hooks are all invoked in a single thread, and initializeExecutionEngineContext is called once and only once.

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/LocalTokenSelectionCriteriaFilter.java
david-stephan: Not in the context of parallel executions, which currently is not fully supported by the local CLI executions, this code should still be ready for it  as the k8s driver

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/NodeLocalAgentProvider.java
david-stephan: thx

step-agent-provisioning-local/src/main/java/step/agents/provisioning/local/ScriptEngineLibraries.java
david-stephan: added a check to nestedParts, but the path is theoretically not reachable

step-cli/step-cli-launcher/src/test/java/step/cli/StepConsoleTest.java
david-stephan: indeed

step-cli/step-cli-launcher/src/main/java/step/cli/StepConsole.java
david-stephan: refactored

step-plans/step-plans-base-artefacts/src/main/java/step/engine/plugins/FunctionPlugin.java
david-stephan: comment updated

step-cli/step-cli-launcher/src/test/java/step/cli/StepConsoleTest.java
```diff
@@ -322,7 +322,7 @@ public void testLocalExecute() {
         Histories histories = new Histories(null, null, null, localExecuteHistory);
 
         // all parameters
-        int res = runMain(histories, "ap", "execute", "-p=src/test/resources/samples/step-automation-packages-sample1.jar", "--local", "--includePlans=p1,p2", "--excludePlans=p3,p4", "--includeCategories=CatA,CatB", "--excludeCategories=CatC,CatD", "-ep=key1=value1|key2=value2", "-ep=key3=value3");
+        int res = runMain(histories, "ap", "execute", "-p=src/test/resources/samples/step-automation-packages-sample1.jar", "--local", "--includePlans=p1,p2", "--excludePlans=p3,p4", "--includeCategories=CatA,CatB", "--excludeCategories=CatC,CatD", "-ep=key1=value1|key2=value2", "-ep=key3=value3", "--localAgentVmArgs=-Xmx4g", "--localAgentVmArgs=-XX:HeapDumpPath=C:\\Program Files\\dumps");
```
cl-exense, line 370: This is a bad idea and is not platform-agnostic. It uses paths that will not be present on most systems (including windows, I don't know how it would behave there), and it would create a leftover directory (literally named "C:\Program Files\dumps") on non-windows machines. I assume that under normal circumstances we don't expect this directory to be used because we don't expect heap dumps, but still...

If this is absolutely required under some circumstances for debugging or so, I suggest to keep the relevant definitions as comments so they can easily be enabled when needed - but not as a default definition that will be used on normal test runs.

step-cli/step-cli-launcher/src/test/java/step/cli/StepConsoleTest.java
```diff
@@ -333,6 +333,8 @@ public void testLocalExecute() {
         Assert.assertEquals("CatC,CatD", usedParams.excludeCategories);
         Assert.assertEquals(Map.of("key1", "value1", "key2", "value2", "key3", "value3"), usedParams.executionParameters);
         Assert.assertEquals("step-automation-packages-sample1.jar", usedParams.apFile.getName());
+        Assert.assertEquals(List.of("-Xmx4g", "-XX:HeapDumpPath=C:\\Program Files\\dumps"),
```
cl-exense, line 381: Related to above comment.

If the purpose of the argument is actually purely for tests, then simply use a relative directory name like "./threaddumps" that's not OS-specific.

step-cli/step-cli-launcher/src/test/java/step/cli/StepConsoleTest.java
cl-exense: Hehehe, thanks for the acknowledgement. Then why keep it? ;-)

cl-exense (changes requested):
Two small nitpicks remaining.

step-cli/step-cli-launcher/src/test/java/step/cli/StepConsoleTest.java
david-stephan: This works anywhere: 1. This test does not cover actual execution which is mocked, 2. the VM arg is HeapDumpPath which does nothing unless a heap dump is triggered. This test covers among other things that such arguments (including backslahes and spaces) are properly handled.

step-cli/step-cli-launcher/src/test/java/step/cli/StepConsoleTest.java
david-stephan: see previous comment

step-cli/step-cli-launcher/src/test/java/step/cli/StepConsoleTest.java
david-stephan: Well I did not just all occurrences :-), updated now

step-cli/step-cli-launcher/src/test/java/step/cli/StepConsoleTest.java
cl-exense: In that case, fine with me.

cl-exense (dismissed):

cl-exense (dismissed):
Loked over the latest changes; while I trust you to better understand the full details, I don't see any glaring problems.

PR #693, opened 2026-08-19
Sed 4841 automation package file system resource resolver

david-stephan:


step-core/src/main/java/step/automation/packages/ApResourceMaterializer.java
```diff
@@ -0,0 +1,201 @@
...
+                atomicMove(tmp, target.toPath());
+            } finally {
+                // No-op on the success path: atomicMove has renamed tmp onto target, so it no longer
+                // exists. This only removes a stray temp tree left behind when the copy or move threw.
+                if (Files.exists(tmp)) {
+                    FileHelper.deleteFolder(tmp.toFile());
+                }
+            }
+        }
+    }
```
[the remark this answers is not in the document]
david-stephan: addressed

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/AutomationPackageExecutionPlugin.java
```diff
@@ -68,6 +77,24 @@ public void initializeExecutionContext(ExecutionEngineContext executionEngineCon
...
+     * context's own {@link step.attachments.FileResolver}, and the concrete archive type is chosen by
+     * the reader registry.
+     */
+    private void installApResourceProvider(ExecutionContext executionContext) {
+        if (apResourceCacheRoot == null || automationPackageReaderRegistry == null) {
+            return;
+        }
+        executionContext.setApResourceProvider(new AutomationPackageResourceProvider(
+            apResourceCacheRoot,
+            () -> executionContext.get(AutomationPackageAccessor.class),
```
jeromecomte, line 93: I guess we're doing this because it isn't available when installApResourceProvider is called. Can't we ensure this with plugin dependencies? 
if the accessor isn't optional, shouldn't we call reuire instead?

step-automation-packages/step-automation-packages-junit-core/src/main/java/step/engine/plugins/AutomationPackageLocalOSPlugin.java
```diff
@@ -67,6 +71,24 @@ public void initializeExecutionEngineContext(AbstractExecutionEngineContext pare
...
+                createApResourceCacheRoot(),
+                automationPackageManager::getAutomationPackageAccessor,
+                archiveReference -> context.getFileResolver().resolve(archiveReference),
+                file -> automationPackageReaderRegistry.getReaderForFile(file).createAutomationPackageArchive(file, null, null)));
+        }
+    }
+
+    private static File createApResourceCacheRoot() {
+        try {
+            return Files.createTempDirectory("ap-cache-local").toFile();
```
jeromecomte: Not able to find where this temporary directory is deleted.

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/AutomationPackageManager.java
```diff
@@ -97,6 +97,14 @@ public class AutomationPackageManager {
     private final int maxParallelVersionsPerPackage;
     private final ObjectHookRegistry objectHookRegistry;
 
+    /**
+     * Materialisation root for {@code apResource:} files. Set only on the MAIN manager (see
```
jeromecomte: Don't fully understand which path this cache covers? does it only apply for deployed APs on a server instance?

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/AutomationPackagePlansAttributesApplier.java
```diff
@@ -150,13 +149,9 @@ private void applyResourcePropertyRecursively(Object object, StagingAutomationPa
...
-        AutomationPackageResourceUploader resourceUploader = new AutomationPackageResourceUploader();
-        Resource resource = resourceUploader.uploadResourceFromAutomationPackage(yamlResourceRef, ResourceManager.RESOURCE_TYPE_DATASOURCE, apContext);
-        String result = null;
-        if (resource != null) {
-            result = FileResolver.RESOURCE_PREFIX + resource.getId().toString();
-        }
-        return result;
+        // Use the context's uploader so the rewrite (server) vs pass-through (IDE) behaviour matches
+        // the keyword plugins. It rewrites a plain archive-relative path into an apResource: reference.
+        return apContext.getResourceUploader().applyResourceReference(yamlResourceRef, ResourceManager.RESOURCE_TYPE_DATASOURCE, apContext);
```
jeromecomte: Do we still need a resource type now that we don't create resources?

step-core/src/main/java/step/automation/packages/ApResourceBrowser.java
```diff
@@ -0,0 +1,563 @@
...
+ * the picker offers exactly what the CLI would put into the package. A zip is not: the CLI applies the
+ * very same filter when it builds it, so its ignored entries are already gone, and interpreting a
+ * {@code .apignore} that happens to sit inside an archive would hide entries that <i>are</i> part of
+ * that package.
+ * <p>
+ * Nothing is materialised into the {@code apResource:} cache here — that cache is owned by the
+ * execution path ({@link ApResourceMaterializer}) and a UI browse or download must not pollute it,
+ * nor extract a whole directory just to answer a listing.
+ */
+public class ApResourceBrowser {
```
jeromecomte, line 74: This is a lot of code for something that seems to be quite "standard". Just wondering if there are no libraries for this...

step-core/src/main/java/step/automation/packages/ApResourceMaterializer.java
```diff
@@ -0,0 +1,178 @@
...
+     *                        materialised.
+     * @return the materialised file (or directory), never {@code null}
+     * @throws ApResourceNotFoundException if the entry is absent from the archive
+     */
+    public File materialize(File cacheRoot, String apId, String relativePath, Supplier<AutomationPackageArchive> archiveSupplier) {
+        Objects.requireNonNull(cacheRoot, "cacheRoot must not be null");
+        Objects.requireNonNull(apId, "apId must not be null");
+        Objects.requireNonNull(relativePath, "relativePath must not be null");
+        Objects.requireNonNull(archiveSupplier, "archiveSupplier must not be null");
+        String normalized = FileResolver.normalizeApRelativePath(relativePath);
```
jeromecomte, line 76: This normalize path method is called at many different places. Just wondering if we couldn't call it at a more central place (at the interfaces) in order not having to deal with normalization later

step-core/src/main/java/step/automation/packages/ApResourceMaterializer.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    private void materializeFile(URL url, File target) throws IOException {
+        Path parent = target.toPath().getParent();
+        Path tmp = Files.createTempFile(parent, TMP_PREFIX, ".tmp");
+        try {
+            try (InputStream in = openStreamWithoutCaching(url)) {
+                Files.copy(in, tmp, StandardCopyOption.REPLACE_EXISTING);
+            }
+            atomicMove(tmp, target.toPath());
```
jeromecomte: Why do we need this if we anyway hold a lock on target?

step-core/src/main/java/step/automation/packages/ApResourceMaterializer.java
```diff
@@ -0,0 +1,178 @@
...
+     * The content is extracted straight into the temporary sibling of the target rather than into a
+     * temporary directory of its own that would then have to be copied over: a directory resource is
+     * a whole subtree, and walking and writing it twice is worth avoiding.
+     */
+    private void materializeDirectory(URL url, File target) throws Exception {
+        Path parent = target.toPath().getParent();
+        Path tmp = Files.createTempDirectory(parent, TMP_PREFIX);
+        try {
+            ClassLoaderResourceFilesystem.extractDirectory(url, tmp);
+            atomicMove(tmp, target.toPath());
```
jeromecomte: Idem

step-core/src/main/java/step/automation/packages/AutomationPackageResourceMapper.java
```diff
@@ -45,87 +43,33 @@ public String applyUniqueResourceReference(String resourceReference,
...
-                    } catch (IOException e) {
-                        logger.error("Unable to close the automation package input stream.");
-                    }
-                }
-                if (tempFolder != null) {
-                    // Delete the temporary folder
-                    FileHelper.deleteFolder(tempFolder);
-                }
-            }
+        if (FileResolver.isResource(resourceReference) || FileResolver.isApResource(resourceReference)) {
```
jeromecomte: In which case do we already have an ApResource here?

step-core/src/main/java/step/automation/packages/ClassLoaderResourceFilesystem.java
```diff
@@ -162,16 +204,48 @@ private static URL toURL(URI uri) {
...
 
         public JarResourcePath(URL url) throws MalformedURLException {
+            // The file of a jar URL looks like "file:/path/to/my%20jar.jar!/my%20folder/my%20file.txt".
+            // Both parts are percent encoded and have to be decoded before being used as a filesystem
+            // path, respectively as a zip entry name.
             String urlFile = url.getFile();
             int bangIndex = urlFile.indexOf('!');
-            pathInJar = urlFile.substring(bangIndex + 2);
-            jarFile = new URL(urlFile.substring(0, bangIndex)).getFile();
+            pathInJar = decodePath(urlFile.substring(bangIndex + 2));
```
jeromecomte: On branch 4811 I could fix this without explicit decode using: url.toURI().getSchemeSpecificPart(); means:

public JarResourcePath(URL url) throws MalformedURLException, URISyntaxException {
            // use url.toURI().getSchemeSpecificPart() instead of url.getPath() in order to get the unencoded path (%20 replaced by " ")
            String urlFile = url.toURI().getSchemeSpecificPart();
            int bangIndex = urlFile.indexOf('!');
            pathInJar = urlFile.substring(bangIndex + 2);
            jarFile = new URL(urlFile.substring(0, bangIndex)).getFile();
        }

step-plans/step-plans-base-artefacts/src/main/java/step/datapool/excel/ExcelFileLookup.java
```diff
@@ -40,7 +40,12 @@ public ExcelFileLookup(ExecutionContext context) {
 
     public File lookup(String workbookPath) {
         File workBookFile;
-        if (workbookPath.contains("/") || workbookPath.contains("\\")) {
+        if (fileResolver != null && (FileResolver.isResource(workbookPath) || FileResolver.isApResource(workbookPath))) {
```
jeromecomte: Shouldn't we throw an error if the fileResolver is null an the workbookPath is a resource?

jeromecomte: We're already doing the same in the else block

jeromecomte (changes requested):
Overall, the PR looks good.
Didn't look at the JUnit test logic in detail but the coverage looks good too.
No major finding. 
Noted some minor points.

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/AutomationPackageExecutionPlugin.java
david-stephan: For isolated execution, the layered accessor is created and added to  the context  in importArtefact of the related AP repository which is done after the initializeExecutionContext hook

step-automation-packages/step-automation-packages-junit-core/src/main/java/step/engine/plugins/AutomationPackageLocalOSPlugin.java
david-stephan: fixed

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/AutomationPackageManager.java
david-stephan: updated comment.

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/AutomationPackagePlansAttributesApplier.java
david-stephan: no, refactored

step-core/src/main/java/step/automation/packages/ApResourceBrowser.java
david-stephan: not really, zipfs could replace a part of the code but keeps a handle on the archive blocking clean up on windows and not worth a change

step-core/src/main/java/step/automation/packages/ApResourceMaterializer.java
david-stephan: We only hold the lock for writing, reader return directly without locking if the target exists, this the target must only exists when the content is fully materialized, this also prevent corrupted/partial extraction to ever reach the cache

step-core/src/main/java/step/automation/packages/ApResourceMaterializer.java
david-stephan: idem

step-core/src/main/java/step/automation/packages/ApResourceMaterializer.java
david-stephan: I don't see a common central place for these usages

step-core/src/main/java/step/automation/packages/ClassLoaderResourceFilesystem.java
david-stephan: refactored

step-plans/step-plans-base-artefacts/src/main/java/step/datapool/excel/ExcelFileLookup.java
david-stephan: refactored to address both comment. The only require change with the new apResource is how to check if its's a file system path, because apResource may also contains slashes

step-core/src/main/java/step/automation/packages/ApResourceMaterializer.java
```diff
@@ -0,0 +1,178 @@
...
+            Files.move(source, target, StandardCopyOption.ATOMIC_MOVE);
+        } catch (AtomicMoveNotSupportedException e) {
+            // Defensive fallback. We always create the temp in the target's own parent directory, so
+            // source and target share a filesystem and the usual cause of this exception (a
+            // cross-filesystem move, EXDEV) cannot occur here. It can still be thrown by exotic
+            // java.nio providers that simply don't implement atomic move (some FUSE/overlay or
+            // network filesystems), in which case a plain same-filesystem move (rename) is used.
+            Files.move(source, target);
+        }
+    }
```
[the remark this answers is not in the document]
david-stephan: implemented to be more robust, but the case could never happen within the same JVM as a lock is used on the targer.

step-functions-plugins/step-functions-plugins-java/step-functions-plugins-java-def/src/main/java/step/plugins/java/ApScriptFileWriter.java
```diff
@@ -0,0 +1,137 @@
...
+                LocalApResourceWriter.createFileIfMissing(automationPackageRoot, relativePath, templateStream);
+            }
+            function.getScriptFile().setValue(FileResolver.createPathForLocalApResource(relativePath));
+            return automationPackageRoot.resolve(relativePath).toFile();
+        } catch (IOException | IllegalArgumentException e) {
+            throw new SetupFunctionException("Unable to create the script of the keyword '"
+                + function.getAttributes().get(AbstractOrganizableObject.NAME) + "' in the automation package "
+                + automationPackageRoot, e);
+        }
+    }
```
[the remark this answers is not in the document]
david-stephan: implemented

step-automation-packages/step-automation-packages-collections/src/test/java/step/core/collections/AutomationPackageFragmentReferenceTest.java
```diff
@@ -81,10 +81,10 @@ public void testAddCompositeKeywordToNewFragmentAndRenameAndRemove() throws IOEx
 
         assertFilesEqual(
             expectedFilesPath
-                .resolve("Hello World Composite Function.yml"),
+                .resolve("Hello_World_Composite_Function.yml"),
```
jeromecomte, line 84: Don't we support spaces in fragments anymore?

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/AutomationPackageServices.java
```diff
@@ -912,7 +912,11 @@ public String getApResourceReference(@QueryParam("apId") String apId, @QueryPara
         // the path is not required to exist (yet), but it must be a valid archive relative path
         String normalizedPath = normalizeApRelativePath(relativePath);
         // the package must be visible in the current context, even though nothing of it is read here
-        getAutomationPackage(apId);
+        if (FileResolver.LOCAL_AP_ID.equals(apId)) {
+            currentlyEditedAutomationPackageDirectory();
```
jeromecomte, line 916: This is a detail but it feels a bit unnatural to use a get method just for validation (as we don't do anything with its result)

step-automation-packages/step-automation-packages-junit-core/src/main/java/step/engine/plugins/LocalApResourceCacheRoot.java
```diff
@@ -0,0 +1,99 @@
...
+            return new LocalApResourceCacheRoot(Files.createTempDirectory(DIRECTORY_PREFIX).toFile());
+        } catch (IOException e) {
+            throw new RuntimeException("Unable to create the local apResource cache directory", e);
+        }
+    }
+
+    protected LocalApResourceCacheRoot(File root) {
+        this.root = root;
+        this.shutdownHook = new Thread(this::delete, "ap-cache-local-cleanup");
+        Runtime.getRuntime().addShutdownHook(shutdownHook);
```
jeromecomte: Why do we have to register a shutdown hook if it is closed by the engine via its close method()?

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/mappers/CompositeFunctionToYamlMapper.java
```diff
@@ -44,7 +46,12 @@ public YamlAutomationPackageKeyword toYamlObject(CompositeFunction compositeFunc
 
         Plan plan = compositeFunction.getPlan();
         if (plan != null) {
-            YamlPlan yamlPlan = planReader.planToYamlPlan(compositeFunction.getPlan());
+            // As for plans, AP resource references in the composite's plan are transformed back to relative path
+            YamlPlan yamlPlan;
+            try (ResourceReferences.Restoration restoration =
```
jeromecomte: We're not doing anything with the result "restoration". Is this intended? Are the references updated directly in the plan?

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/model/AbstractAutomationPackageFragmentYaml.java
```diff
@@ -159,13 +160,33 @@ public void writeToDisk() {
...
             throw new AutomationPackageWriteToDiskException(MessageFormat.format("Error when writing automation package fragment {0} back to disk.", path), e);
         }
     }
 
+    /**
+     * The {@code apResource:local:} form belongs to the editor's memory, never to the descriptor: each
+     * keyword plugin maps it back in {@code setDeclaredFieldsFromObject}, and the data sources of a plan
+     * in {@code AutomationPackageLocalResourceMapper.toDescriptorReferences}. A plugin that forgets to
+     * would otherwise write an internal reference into a file the user owns and commits, so the write is
+     * refused instead - loudly, and before the file is touched.
```
jeromecomte: Not sure if this is the best strategy to handle such errors. The question is: what is worst for the user? persisting something with a wrong reference or not persisting at all (assuming the user cannot fix this error)?
Would replace by a clear warning for the user and a clear recommendation (report the issue)

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
```diff
@@ -175,7 +177,16 @@ public synchronized <BO extends AbstractOrganizableObject, YO extends PatchableY
         if (mapper == null) {
             throw new AutomationPackageUpdateException("No BusinessObjectToYamlMapper registered for class: " + object.getClass().getName());
         }
-        YO newYamlObject = mapper.toYamlObject(object);
+        YO newYamlObject;
+        // TODO ugly implementation to transform back apResources to pure string path in YAML for resource references in Plans
+        if (object instanceof Plan) {
+            try (ResourceReferences.Restoration restoration =
```
jeromecomte: Idem here. We don't use the result

step-core/src/main/java/step/automation/packages/AutomationPackageResourceMapper.java
```diff
@@ -55,11 +58,20 @@ public String applyUniqueResourceReference(String resourceReference,
...
                                          StagingAutomationPackageContext context) {
         if (resourceReference == null || resourceReference.isEmpty()) {
             return null;
         }
+        if (FileResolver.isLocalApResource(resourceReference)) {
+            // The editor form never reaches this far: the YAML holds plain relative paths, and
+            // AutomationPackageYamlFragmentManager.save strips the prefix before writing. Getting one
+            // here means an entity was staged straight out of the editor, and returning it untouched
+            // as the branch below does would deploy a reference nothing can resolve.
+            throw new RuntimeException("The reference " + resourceReference + " of the automation package "
```
jeromecomte: If this message should reach the users (which seems to be the case), I would probably add a recommendation on how to solve the issue

step-core/src/main/java/step/automation/packages/ResourceReferences.java
```diff
@@ -0,0 +1,215 @@
...
+ * {@link EntityReference}{@code (type = }{@link EntityConstants#resources}{@code )}. That declaration is
+ * the only thing distinguishing a relative path from any other string, and it is what a deployment
+ * follows, so the editor follows exactly the same one - no more, no less.
+ * <p>
+ * Keywords are deliberately <b>not</b> handled here: each plugin maps its own fields in
+ * {@code Yaml*Function.fillDeclaredFields} and back in {@code setDeclaredFieldsFromObject}, because
+ * some of them are conditional. {@code YamlK6Function} builds three different shapes out of two fields,
+ * and only that class can tell which one it produced.
+ */
+public class ResourceReferences {
```
jeromecomte: Not sure how we did this for exports, but I have the feeling we had a more generic implementation. This class (as commented) is specific to plans and handles introspection in its own way. If there's something in common with import/exports and vistors we could reuse, that would be nice

step-core/src/test/java/step/automation/packages/ClassLoaderResourceFilesystemTest.java
jeromecomte, line 1: Skipping this class in review

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
```diff
@@ -175,7 +177,16 @@ public synchronized <BO extends AbstractOrganizableObject, YO extends PatchableY
         if (mapper == null) {
             throw new AutomationPackageUpdateException("No BusinessObjectToYamlMapper registered for class: " + object.getClass().getName());
         }
-        YO newYamlObject = mapper.toYamlObject(object);
+        YO newYamlObject;
+        // TODO ugly implementation to transform back apResources to pure string path in YAML for resource references in Plans
```
jeromecomte: As discussed we could move this to YamlFileDataSource or a more specific place than this one

step-core/src/main/java/step/automation/packages/ResourceReferences.java
jeromecomte, line 1: As discussed, this class could be probably removed as we're already doing such an instrospection for DataSources

step-functions-plugins/step-functions-plugins-java/step-functions-plugins-java-def/src/main/java/step/plugins/java/AbstractScriptFunctionType.java
```diff
@@ -132,88 +140,157 @@ protected String getScriptLanguage(GeneralScriptFunction conf) {
...
-        } catch (FileNotFoundException e) {
-            throw new SetupFunctionException("Unable to apply template. The file '" + templateScript.getAbsolutePath() + "' doesn't exist");
         }
-    }
-
-    protected File setupScriptFile(GeneralScriptFunction function, String templateFilename) throws SetupFunctionException {
-        return setupScriptFile(function, getTemplateFileInputStream(templateFilename));
+        // The distribution's template takes precedence - it is the one an administrator can edit - and the
+        // bundled copy takes over where there is no controller directory to hold one, i.e. in the IDE.
+        InputStream bundledTemplate = AbstractScriptFunctionType.class.getResourceAsStream(BUNDLED_TEMPLATES_PATH + templateFilename);
```
jeromecomte, line 156: I guess we'll have to check that we bundle all templates

step-functions-plugins/step-functions-plugins-java/step-functions-plugins-java-def/src/main/java/step/plugins/java/AbstractScriptFunctionType.java
```diff
@@ -132,88 +140,157 @@ protected String getScriptLanguage(GeneralScriptFunction conf) {
...
+     *                       against a script that exists already leaves it unread
+     */
     protected File setupScriptFile(GeneralScriptFunction function, InputStream templateStream,
                                    String scriptDir) throws SetupFunctionException {
-        File scriptFile;
+        try (InputStream template = templateStream) {
+            File scriptFile;
+
+            Path automationPackageRoot = getEditableAutomationPackageRoot();
+            if (automationPackageRoot != null) {
```
jeromecomte, line 178: Initially this method (setupScriptFile) was intended to create a file on the controller directly. Theoretically it shouldn't be called in the context of the IDE

jeromecomte (changes requested):

step-automation-packages/step-automation-packages-collections/src/test/java/step/core/collections/AutomationPackageFragmentReferenceTest.java
david-stephan: We do, simply when we create a file path based on the entity name (e.g. create a plan, a keyword, the script of a script keyword...) the sanitation now use "_" in place of spaces. This was done when centralizing and aligning the code for the new feature to create script files in the AP

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/AutomationPackageServices.java
david-stephan: not perfect, but commented, so we'll keep it for now, the validation could be removed altogether, it's just building a reference in the end.

step-automation-packages/step-automation-packages-junit-core/src/main/java/step/engine/plugins/LocalApResourceCacheRoot.java
david-stephan: IDE can be interrupted with signal, not convinced the  engine is properly closed, this is a safety net

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/mappers/CompositeFunctionToYamlMapper.java
david-stephan: it is used since it is a closeable and used in a try-with-resource block and yes in modify the plan then restore, this is wrong and also related to the comment  "ugly" in AutomationPackageYamlFragmentManager.save, that I refactored it now. So this is removed.

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
david-stephan: Refactored centrally to YamlResourceReference used when serializing the plan to a YAML plan.

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageYamlFragmentManager.java
david-stephan: Same reply, but refactored anyway

step-core/src/main/java/step/automation/packages/ResourceReferences.java
david-stephan: So the code was refactored and is not used anymore when writing the YAML plan. It is kept when reading from YAML, this implementation already existed in AutomationPackagePlansAttributesApplier.java‎ and is now shared with YamlCompositeFunction (that solve a but, YAM composite function could not use file datasources before)

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/model/AbstractAutomationPackageFragmentYaml.java
david-stephan: So the implementation of writing to Yaml and reading from it is asymetric, and it is possible that while reading an AP resource reference is well converted to apResource:<id>:path, but the counter implementation for writing it lacking the conversion (i.e. a new KW type), So, I changed it to automatically repair the reference (convert it to relative path) and just log a warning.

step-core/src/main/java/step/automation/packages/AutomationPackageResourceMapper.java
david-stephan: Yes, refactored the message

step-core/src/main/java/step/automation/packages/AutomationPackageResourceMapper.java
david-stephan: none you're right, it's now modify to throw an error as for isLocalApResource

step-core/src/main/java/step/automation/packages/ResourceReferences.java
david-stephan: As replied on previous comment, this implementation already existed in AutomationPackagePlansAttributesApplier. The visitor are not meant to modify content only to "visit" so for both reasons I would not refactor this now.

step-functions-plugins/step-functions-plugins-java/step-functions-plugins-java-def/src/main/java/step/plugins/java/AbstractScriptFunctionType.java
david-stephan: this path is still used for Oryon keywords

step-functions-plugins/step-functions-plugins-java/step-functions-plugins-java-def/src/main/java/step/plugins/java/AbstractScriptFunctionType.java
david-stephan: increased test coverage

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/model/AbstractAutomationPackageFragmentYaml.java
```diff
@@ -172,19 +179,31 @@ public void writeToDisk() {
...
-            throw new AutomationPackageWriteToDiskException(MessageFormat.format(
-                "Refusing to write the automation package fragment {0}: it holds an editor internal "
-                    + "reference ({1}). This is an internal error - such a reference must be mapped back "
-                    + "to the path it was authored with before the fragment is written.",
-                path, editorReferencePrefix));
+    static String withoutEditorInternalReferences(String yaml, Path fragmentPath) {
+        if (!yaml.contains(EDITOR_REFERENCE_PREFIX)) {
+            return yaml;
         }
+        log.warn("The automation package fragment {} was serialized holding the editor internal reference "
```
jeromecomte, line 202: Nothing specific to this PR and noting to change but a probably a matter of best practices: In such cases, I really prefer the pattern if - else instead of if > return and implicit else

jeromecomte (approved):

step-automation-packages/step-automation-packages-junit-core/src/main/java/step/engine/plugins/LocalApResourceCacheRoot.java
jeromecomte: In this case I would prefer a generic hook to close the engine rather than each plugin having to care about this

step-core/src/main/java/step/automation/packages/ApResourceMaterializer.java
jeromecomte: OK

step-cli/step-cli-launcher/src/main/java/step/cli/GracefulEngineShutdown.java
```diff
@@ -0,0 +1,105 @@
...
+
+    GracefulEngineShutdown(ExecutionEngine engine, long timeoutMs, Consumer<ExecutionContext> abort) {
+        this.engine = Objects.requireNonNull(engine, "engine must not be null");
+        this.timeoutMs = timeoutMs;
+        this.abort = Objects.requireNonNull(abort, "abort must not be null");
+    }
+
+    @Override
+    public void close() {
+        abortRunningExecutions();
```
jeromecomte, line 70: In theory we should first disable the scheduler so that no new execution starts but this is very unlikely in this case. If we would reuse this code for the controller, we should however do this

step-plans/step-plans-yaml-support/src/main/java/step/plans/parser/yaml/model/YamlResourceReference.java
```diff
@@ -102,7 +102,20 @@ public void setResourceId(String resourceId) {
         this.resourceId = resourceId;
     }
 
+    /**
```
jeromecomte, line 105: I assume that the same rule applies when the FE sets an empty reference

step-cli/step-cli-launcher/src/main/java/step/cli/CliShutdownHook.java
```diff
@@ -0,0 +1,127 @@
...
+ * <p>
+ * Resources are closed in reverse registration order, on the assumption that what was built later may
+ * depend on what was built earlier. Note that this orders the <i>registered</i> resources only: it says
+ * nothing about the order in which a resource closes what it owns internally - {@code AbstractContext.close()},
+ * for one, closes the attributes of an execution engine context in unspecified order.
+ * <p>
+ * The hook itself is registered lazily, on the first registration, so that a command owning nothing -
+ * {@code --help}, a remote deployment - installs nothing.
+ */
+public class CliShutdownHook {
```
jeromecomte, line 51: This class is a bit over engineered in my opinion, as we have only on object registering him self

step-plans/step-plans-yaml-support/src/main/java/step/plans/parser/yaml/model/YamlResourceReference.java
david-stephan: yes, it's the same thing/path

step-cli/step-cli-launcher/src/main/java/step/cli/GracefulEngineShutdown.java
david-stephan: I could not find where we would be using the scheduler for the CLI local execution. If we were to integrate this to the controller shutdown sequence, the scheduler should be closed by what creates it (SchedulerPlugin), it's actually already the case.

PR #699, opened 2026-08-28
SED-4912 yaml formatting improve ordering of fields

cmisev:


step-core/src/main/java/step/core/yaml/model/AbstractYamlArtefact.java
```diff
@@ -58,20 +59,28 @@ public abstract class AbstractYamlArtefact<T extends AbstractArtefact> extends A
...
     protected DynamicValue<Boolean> instrumentNode = new DynamicValue<>(false);
     protected DynamicValue<Boolean> continueParentNodeExecutionOnError = new DynamicValue<>(false);
-    protected String description;
 
-    @JsonSchema(ref = YamlJsonSchemaHelper.DEFS_PREFIX + ARTEFACT_ARRAY_DEF)
-    @YamlFieldCustomCopy
-    protected List<NamedYamlArtefact> children = new ArrayList<>();
 
     @YamlFieldCustomCopy
+    @YamlFieldOrder(YamlFieldPriority.LOW)
```
david-stephan, line 74: I haven't tested it, but according to the modify JSON schema on this branch the order is now children -> before -> after instead of before > children > after. Please check it and also if we already have unit tests that validates it.

step-core/src/main/java/step/core/yaml/model/SimpleYamlArtefact.java
```diff
@@ -37,8 +40,8 @@
  */
 public class SimpleYamlArtefact<T extends AbstractArtefact> extends AbstractYamlArtefact<T> {
 
-    @JsonIgnore
-    protected ObjectNode fieldValues;
+    @YamlFieldCustomCopy
```
david-stephan: I assume this has ultimately the same effect than @JsonIgonre, right ?

step-core/src/main/java/step/core/yaml/model/SimpleYamlArtefact.java
```diff
@@ -71,7 +74,18 @@ protected void fillYamlArtefactFields(T artefact) {
 
     public ObjectNode toFullJson() {
         ObjectNode jsonNode = yamlObjectMapper.valueToTree(this);
```
david-stephan, line 76: This seems relatively complex also in term of perforamnce, and previously the setAll would override any existing properties values with the one of the fieldValues object. Not verified, but it seems that @JsonAnyGetter would be more adapted

cmisev: JsonAnyGetter won't control for field ordering, tried that. My honest opinion is that SimpleYAMLArtefact as a whole introduces a lot of complexity and breaks many of the rich features and conventions of the Jackson library, with all the conversions to and from object to JsonObject. Even the avoidance of printing default values in the yaml is perfectly well covered by Jackson annotations and does not require any copying back and forth.

This complexity with the SimpleYamlArtefact is also the primary reason why all my ideas for the diff engine fell flat at the beginning of the year. I'd love to get rid of this altogether, but I always deemed it too much of a can of worms. But maybe with the knowledge I have now it would be of great benefit down the road to simplify it. Let me know if I should tackle this.

step-core/src/main/java/step/core/yaml/model/AbstractYamlArtefact.java
cmisev: The json schema does not seem to enforce ordering, and from what I saw in the generated json the order is completely reverse. Rudimentary covarage is given by the ap collection unit tests but you are right we should add some more. Will do.

cmisev: Hi @david-stephan I added the unit tests

step-core/src/main/java/step/core/yaml/model/SimpleYamlArtefact.java
david-stephan: We know that we have to review and refactor the whole implementation around reading/writing YAML for AP entities in order to algin it, simplify it, make it more pluggable, but not something we achieve for the next major release.

step-core/src/main/java/step/core/yaml/model/AbstractYamlArtefact.java
david-stephan: Thanks, for the test. I guess the order in the schema may impact the auto-completion in intellij, but that's another topic.

david-stephan (approved):

step-core/src/main/java/step/core/yaml/model/SimpleYamlArtefact.java
```diff
@@ -71,7 +74,12 @@ protected void fillYamlArtefactFields(T artefact) {
...
-        jsonNode.setAll(fieldValues);
+        // Flatten fieldValues into existing serialization to preserve order
+        List<Map.Entry<String, JsonNode>> list = jsonNode.propertyStream()
+            .flatMap(e -> e.getKey().equals("fieldValues") ? e.getValue().propertyStream() : Stream.of(e))
+            .toList();
```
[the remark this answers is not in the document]
cmisev: Suggestion declined as it works and is leaner for this purpose.

step-core/src/main/java/step/core/yaml/model/AbstractYamlArtefact.java
cmisev: @david-stephan Good point. Upon doing the other changes of in SED-4905 I had to update the schema anyway, I will then add the changes to the SED-4905 branch once this one is merged

cmisev: @david-stephan I also saw that for the ordering in the schema we need to implement the same logic to the YAMLJsonSchemaHelper. I did not get this at first, will do so for completeness

cmisev:
Hi @david-stephan I have addressed your comments, the schema ordering is shelved, will properly test it locally first, but as discussed yesterday it will be included in the parent task

step-plans/step-plans-base-artefacts/src/main/java/step/artefacts/automation/YamlCallPlan.java
```diff
@@ -29,11 +29,11 @@ public class YamlCallPlan extends AbstractYamlArtefact<CallPlan> {
     private String planId;
 
     @YamlFieldCustomCopy
-    @YamlFieldOrder(YamlFieldPriority.TOP)
+    @YamlFieldOrder(YamlFieldPriority.DESCRIPTORS)
```
david-stephan: for the call keyword we removed those priorities which fall back to the ARTEFACT_SPECIFIC one, if think we should to the same for the callPlan

david-stephan (changes requested):

david-stephan (approved):

PR #700, opened 2026-08-30
SED-4906 Support string interpolation in plan values

jeromecomte:


step-controller/step-controller-server/src/main/java/step/migration/tasks/V31_0_EscapeStringInterpolationInPlanValues.java
```diff
@@ -0,0 +1,98 @@
...
+        }
+    }
+
+    private void escapeCollection(String collectionName) {
+        Collection<Document> collection = collectionFactory.getCollection(collectionName, Document.class);
+        AtomicLong migratedCount = new AtomicLong();
+        AtomicLong errorCount = new AtomicLong();
+
+        logger.info("Escaping the string interpolation placeholders of the plain values in '{}'...", collectionName);
+        try (Stream<Document> documents = collection.findLazy(Filters.empty(), null, null, null, 0)) {
```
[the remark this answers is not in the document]
jeromecomte: The Halloween problem should not occur in MongoDB 5+with WiredTiger. Didn't find any place in the code where we sort by _id, so I prefer let this untouched. @david-stephan what do you think?

step-controller/step-controller-server/src/main/java/step/migration/tasks/StringInterpolationEscaper.java
```diff
@@ -0,0 +1,175 @@
...
+     * @return the escaped document, or null if it isn't a JSON object or if nothing had to be escaped
+     */
+    private static String escapeNestedDocument(String json) {
+        if (json == null || json.isBlank() || !json.contains(InterpolatedString.EXPRESSION_PREFIX)) {
+            return null;
+        }
+        Map<String, Object> parsed;
+        try {
+            parsed = OBJECT_MAPPER.readValue(json, new TypeReference<>() {
+            });
```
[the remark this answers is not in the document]
jeromecomte: This code is only used as part of the migration and only for documents that have to be escaped, which probably represents a very small amount. Keeping as is to avoid new iterations

step-automation-packages/step-automation-packages-manager/src/main/java/step/automation/packages/AutomationPackageReader.java
```diff
@@ -174,15 +174,25 @@ protected AutomationPackageContent newContentInstance() {
...
 
         if (!fragment.getFragments().isEmpty()) {
             for (String importedFragmentReference : fragment.getFragments()) {
                 List<URL> resources = archive.getResourcesByPattern(importedFragmentReference);
                 for (URL resource : resources) {
                     try (InputStream fragmentYamlStream = resource.openStream()) {
-                        fragment = getOrCreateDescriptorReader().readAutomationPackageFragment(fragmentYamlStream, importedFragmentReference, archive.getAutomationPackageName());
-                        fillAutomationPackageWithImportedFragments(targetPackage, fragment, archive);
+                        fragment = getOrCreateDescriptorReader().readAutomationPackageFragment(fragmentYamlStream, importedFragmentReference, archive.getAutomationPackageName(), packageVersion);
+                        fillAutomationPackageWithImportedFragments(targetPackage, fragment, archive, packageVersion);
```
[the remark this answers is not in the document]
jeromecomte: Agree. Changed

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageDescriptorReader.java
```diff
@@ -132,6 +169,57 @@ protected <T extends AutomationPackageFragmentYaml> void logAfterRead(String pac
...
+
+        return yamlObjectMapper.writeValueAsString(migratedDocument);
+    }
+
+    /**
+     * Initializes the migration manager with the migrations of the automation package format
+     */
+    protected MigrationManager initMigrationManager() {
+        MigrationManager migrationManager = new MigrationManager();
+        try (AnnotationScanner annotationScanner = AnnotationScanner.forAllClassesFromClassLoader(AutomationPackageMigration.LOCATION, Thread.currentThread().getContextClassLoader())) {
```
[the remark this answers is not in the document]
jeromecomte: The AutomationPackageDescriptorReader is cached in the AutomationPackageReader which again should be instantiated once in the controller. But it's a valid point and we should check how it behaves, especially in the context of the CLI and Junit runner. Keeping open

step-automation-packages/step-automation-packages-samples/step-automation-packages-sample-echo/src/main/resources/automation-package.yml
```diff
@@ -1,4 +1,4 @@
-schemaVersion: 1.0.0
+version: 1.3.0
```
david-stephan, line 1: why did we update all versions of our test resources, they were written for that version, and since there are no content changes they should still be compatible with that version. If we really want to keep it we should ensure we have tests for all previous version.

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageDescriptorReader.java
```diff
@@ -132,6 +169,57 @@ protected <T extends AutomationPackageFragmentYaml> void logAfterRead(String pac
         }
     }
 
+    /**
+     * Applies the migrations of the automation package format to a descriptor or fragment declaring an older schema
+     * version. This concerns the body of the file itself, the plans it contains are migrated by the yaml plan reader.
```
david-stephan: concretely the plans are part of the body too, not sure what a better world would be.

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageDescriptorReader.java
```diff
@@ -80,15 +97,28 @@ protected Class<? extends AutomationPackageDescriptorYaml> getDescriptorClass()
     }
 
     public AutomationPackageFragmentYaml readAutomationPackageFragment(InputStream yamlFragment, String fragmentName, String packageName) throws AutomationPackageReadingException {
```
david-stephan: no usages found

step-plans/step-plans-base-artefacts/src/main/java/step/plans/parser/yaml/migrations/EscapeStringInterpolationYamlMigrationTask.java
```diff
@@ -0,0 +1,196 @@
...
+ * <p>
+ * <b>The field set below is frozen.</b> It describes where a value sits in the schemas up to 1.2.0, which is closed:
+ * a document declaring such a version can never contain a field introduced later. It must therefore not be extended
+ * when new fields appear - a later schema needing the same treatment gets its own task.
+ * <p>
+ * This first task deliberately covers only the fields where a placeholder is realistically found. The full inventory
+ * of the values reaching a DynamicValue is wider, see the migration documentation.
+ */
+@YamlPlanMigration
+public class EscapeStringInterpolationYamlMigrationTask extends AbstractYamlPlanMigrationTask {
```
david-stephan: We should start using the same convention as for the controller DB migration task V1_3_0_...

step-plans/step-plans-base-artefacts/src/main/java/step/plans/parser/yaml/migrations/V1_3_0_EscapeStringInterpolationYamlMigrationTask.java
```diff
@@ -0,0 +1,196 @@
...
+/**
+ * Since the schema version 1.3.0 the plain values of a plan may embed expressions using the {@code ${...}} placeholder
+ * syntax. Values authored against an earlier schema were used literally, so any of them containing {@code ${} would
+ * change meaning, and would in most cases fail the execution with an unresolvable placeholder.
+ * <p>
+ * This task escapes those values so that a package written against schema 1.2.0 or earlier keeps behaving exactly as
+ * it did. Authors opt into the interpolation by bumping the {@code version} of their descriptor and removing the
+ * escaping where they actually want a placeholder.
+ * <p>
+ * <b>The field set below is frozen.</b> It describes where a value sits in the schemas up to 1.2.0, which is closed:
```
david-stephan, line 47: I had the same issue when reading the javadoc for the DB migration tasks, this is generic for any migration tasks, it's written for the targeted version and should never be modified later, right?

step-controller/step-controller-server/src/main/java/step/migration/tasks/StringInterpolationEscaper.java
```diff
@@ -0,0 +1,178 @@
...
+ * YAML migration, which operate on untyped documents.
+ * <p>
+ * The values of the fields listed in {@link #CONTAINER_FIELDS} are not escaped themselves. They hold a JSON document
+ * (keyword inputs, selection criteria) which isn't interpolated as a whole, see {@link NoStringInterpolation}. Their
+ * content is parsed and the plain values it contains are escaped instead.
+ */
+public class StringInterpolationEscaper {
+
+    /**
+     * The fields holding a JSON document rather than user facing text, as of the schema this migration applies to.
```
david-stephan: The comment is confusing. What "shema"? I would warn that this class is used by the migration task, V31_0_EscapeStringInterpolationInPlanValues and as any migration task it shall not be modified for future releases

step-controller/step-controller-server/src/main/java/step/migration/tasks/StringInterpolationEscaper.java
```diff
@@ -0,0 +1,178 @@
...
+import step.core.dynamicbeans.DynamicValue;
+import step.core.dynamicbeans.InterpolatedString;
+import step.core.dynamicbeans.NoStringInterpolation;
+
+/**
+ * Escapes the plain values of a document authored before the string interpolation existed, so that they keep being
+ * used literally. See {@link InterpolatedString#escape(String)} for the escaping itself.
+ * <p>
+ * The document is walked generically: any nested map shaped like a serialized {@link DynamicValue} holding a plain
+ * string value is escaped, wherever it sits in the structure. This is used by both the database migration and the
```
david-stephan: Comment seems wrong, I only see usages in 31_0_EscapeStringInterpolationInPlanValues

step-controller/step-controller-server/src/main/java/step/migration/tasks/StringInterpolationEscaper.java
```diff
@@ -0,0 +1,178 @@
...
+        String value = (String) map.get(VALUE_FIELD);
+        if (isContainer) {
+            String escapedDocument = escapeNestedDocument(value);
+            if (escapedDocument != null) {
+                map.put(VALUE_FIELD, escapedDocument);
+                return true;
+            }
+            return false;
+        }
+        String escaped = InterpolatedString.escape(value);
```
david-stephan, line 132: else block ?

step-controller/step-controller-server/src/main/java/step/migration/tasks/StringInterpolationEscaper.java
```diff
@@ -0,0 +1,178 @@
...
+        try {
+            parsed = OBJECT_MAPPER.readValue(json, new TypeReference<>() {
+            });
+        } catch (Exception e) {
+            // Not a JSON object. The value is left untouched: it isn't interpolated as a whole either
+            return null;
+        }
+        if (!escapeMap(parsed, true)) {
+            return null;
+        }
```
david-stephan: else block ?

step-controller/step-controller-server/src/main/java/step/migration/tasks/StringInterpolationEscaper.java
```diff
@@ -0,0 +1,178 @@
...
+ * <p>
+ * The document is walked generically: any nested map shaped like a serialized {@link DynamicValue} holding a plain
+ * string value is escaped, wherever it sits in the structure. This is used by both the database migration and the
+ * YAML migration, which operate on untyped documents.
+ * <p>
+ * The values of the fields listed in {@link #CONTAINER_FIELDS} are not escaped themselves. They hold a JSON document
+ * (keyword inputs, selection criteria) which isn't interpolated as a whole, see {@link NoStringInterpolation}. Their
+ * content is parsed and the plain values it contains are escaped instead.
+ */
+public class StringInterpolationEscaper {
```
david-stephan, line 42: The unit test seems to be covering also error cases, but should be tested with complex plans with all possible constellation of controls, values and expressions.

step-plans/step-plans-base-artefacts/src/main/java/step/plans/parser/yaml/migrations/V1_3_0_EscapeStringInterpolationYamlMigrationTask.java
```diff
@@ -0,0 +1,196 @@
...
+ * <p>
+ * This first task deliberately covers only the fields where a placeholder is realistically found. The full inventory
+ * of the values reaching a DynamicValue is wider, see the migration documentation.
+ */
+@YamlPlanMigration
+public class EscapeStringInterpolationYamlMigrationTask extends AbstractYamlPlanMigrationTask {
+
+    private static final Logger logger = LoggerFactory.getLogger(EscapeStringInterpolationYamlMigrationTask.class);
+
+    private static final String ROOT = "root";
```
david-stephan, line 59: The 4 constants could probably be moved to the parent

step-controller/step-controller-server/src/main/java/step/migration/tasks/V31_0_EscapeStringInterpolationInPlanValues.java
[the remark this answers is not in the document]
david-stephan: Should not impact this query since we don't modify the index or use unique index. According to the doc :
Ref: https://www.mongodb.com/docs/manual/core/read-isolation-consistency-recency/?language-no-dependencies=python#cursor-snapshot
Cursor Snapshot
MongoDB cursors can return the same document more than once in some situations. As a cursor returns documents, other operations may interleave with the query. If one of these operations changes the indexed field on the index used by the query, then the cursor could return the same document more than once.

Queries that use unique indexes can, in some cases, return duplicate values. If a cursor using a unique index interleaves with a delete and insert of documents sharing the same unique value, the cursor may return the same unique value twice from different documents.

step-controller/step-controller-server/src/main/java/step/migration/tasks/V31_0_EscapeStringInterpolationInPlanValues.java
```diff
@@ -0,0 +1,97 @@
...
+            documents.forEach(document -> {
+                try {
+                    if (StringInterpolationEscaper.escapeDocument(document)) {
+                        collection.save(document);
+                        migratedCount.incrementAndGet();
+                        logger.info("Escaped the plain values of the entry {} of '{}'", document.getId(), collectionName);
```
david-stephan, line 76: do we want a log entry for each DB records ?

step-core/src/main/java/step/core/dynamicbeans/InterpolatedString.java
```diff
@@ -0,0 +1,220 @@
...
+     */
+    public static InterpolatedString parse(String source) {
+        if (source == null) {
+            throw new IllegalArgumentException("The string to be parsed must not be null");
+        }
+        InterpolatedString cached = parseCache.getIfPresent(source);
+        if (cached != null) {
+            return cached;
+        }
+        // Malformed strings aren't cached. They are rare and are re-parsed on every evaluation
```
david-stephan: comment is a bit weird, next lines does the parsing (always)  but the cache is only updated if no StringInterpolationException is thrown

step-core/src/main/java/step/core/dynamicbeans/InterpolatedString.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+
+    private static InterpolatedString doParse(String source) {
+        List<Segment> segments = new ArrayList<>();
+        StringBuilder literal = new StringBuilder();
+        int literalOffset = 0;
+        int length = source.length();
+        int i = 0;
+        while (i < length) {
```
david-stephan, line 179: Feels a bit old school, but I guess it's fine

step-core/src/main/java/step/core/dynamicbeans/StringInterpolator.java
```diff
@@ -0,0 +1,131 @@
...
+                Object o = expressionHandler.evaluateGroovyExpression(segment.getText(), bindings, hasProtectedAccess);
+                if (o instanceof ProtectedVariable) {
+                    // The clear value of a protected variable is only rendered into the result returned to
+                    // callers with protected access. It is never rendered into the obfuscated result
+                    ProtectedVariable protectedVariable = (ProtectedVariable) o;
+                    obfuscatedValue.append(protectedVariable.obfuscatedValue);
+                    clearValue.append(hasProtectedAccess ? String.valueOf(protectedVariable.value) : protectedVariable.obfuscatedValue);
+                    containsProtectedValues = true;
+                } else {
+                    // Covers GString results as well, whose toString() renders the interpolated value
```
david-stephan: we are in the else block of ProtectedVariable, so I'm not sure I get the comment

step-core/src/main/java/step/core/dynamicbeans/DynamicValueResolver.java
```diff
@@ -26,45 +26,82 @@
...
         if (dynamicValue.isDynamic()) {
-            // TODO support different expression types
-            String exprType = dynamicValue.expressionType;
-            EvaluationResult result = new EvaluationResult();
-            try {
-                Object evaluationResult;
-                Object protectedResult = null;
+            dynamicValue.evalutationResult = getEvaluationResult(dynamicValue.expression, bindings, dynamicValue.hasProtectedAccess());
+        } else {
+            // A null result means that the value requires no interpolation and is to be returned as is
```
david-stephan: comment is not clear or not adapted for the next line of code

step-core/src/main/java/step/core/dynamicbeans/DynamicValueResolver.java
```diff
@@ -26,45 +26,82 @@
...
+            if (interpolation == null) {
+                return null;
+            }
+            // The obfuscated value is identical to the clear one unless a protected variable was interpolated
+            result.setResultValue(interpolation.getObfuscatedValue());
+            result.setProtectedValue(interpolation.getValue());
+        } catch (Exception e) {
+            result.setEvaluationException(e);
+        }
+        return result;
```
david-stephan, line 74: I would expect to return null when there is nothing to interpolate, something is not clear to me.

david-stephan (commented):
The implementation of the feature looks good with some comments. For the migration part is looks complex and more difficult to assess the risks. Also seeing the size of the migration notes might scare / confuse some users. I wonder, given that values with such syntax  "${" are probably very rare, if we should not have just notified users in the migration notes and add a step property to opt out (i.e. disable interpolation) ....

step-automation-packages/step-automation-packages-samples/step-automation-packages-sample-echo/src/main/resources/automation-package.yml
jeromecomte: @david-stephan All these files where using the field schemaVersion which doesn't exist anymore. If version is not set, the version falls back to the current one. I therefore decided to pin them to 1.3.0 for the future. 1.2.0 would have also worked

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageDescriptorReader.java
jeromecomte: agree. adapted

step-automation-packages/step-automation-packages-yaml/src/main/java/step/automation/packages/yaml/AutomationPackageDescriptorReader.java
jeromecomte: you're right. removed

step-plans/step-plans-base-artefacts/src/main/java/step/plans/parser/yaml/migrations/EscapeStringInterpolationYamlMigrationTask.java
jeromecomte: OK. changed

step-plans/step-plans-base-artefacts/src/main/java/step/plans/parser/yaml/migrations/V1_3_0_EscapeStringInterpolationYamlMigrationTask.java
jeromecomte: Exact. These fields corresponds to the fields that are existing in the schema 1.2.0 and that we migrate. They should not be modified later

step-controller/step-controller-server/src/main/java/step/migration/tasks/StringInterpolationEscaper.java
jeromecomte: Agree. Improved

step-controller/step-controller-server/src/main/java/step/migration/tasks/StringInterpolationEscaper.java
jeromecomte: Fixed

step-controller/step-controller-server/src/main/java/step/migration/tasks/StringInterpolationEscaper.java
jeromecomte: I checked the coverage and added specific unit tests. I also re-read the code in detail. I am now confident that this codes works as expected

step-plans/step-plans-base-artefacts/src/main/java/step/plans/parser/yaml/migrations/V1_3_0_EscapeStringInterpolationYamlMigrationTask.java
jeromecomte: These are specific to the current state and are frozen. If prefer not moving them to the parent which is not version specific

step-controller/step-controller-server/src/main/java/step/migration/tasks/V31_0_EscapeStringInterpolationInPlanValues.java
jeromecomte: Usually I would say no but here I don't expect a lot of plans to be concerned and would therefore prefer to keep this

step-core/src/main/java/step/core/dynamicbeans/InterpolatedString.java
jeromecomte: you're right. removed

step-core/src/main/java/step/core/dynamicbeans/InterpolatedString.java
jeromecomte: agree but I won't retouch it ;)

step-core/src/main/java/step/core/dynamicbeans/StringInterpolator.java
jeromecomte: the comment has its origin in step.core.dynamicbeans.DynamicValueResolver#convertResultIfRequired. But as we anyway need a string here and have to call toString, this comment doesn't make sense here. I removed it

step-core/src/main/java/step/core/dynamicbeans/DynamicValueResolver.java
jeromecomte: I tried to improve it

step-core/src/main/java/step/core/dynamicbeans/DynamicValueResolver.java
jeromecomte: this is the case at ligne 65

david-stephan (dismissed):

david-stephan (approved):

PR #706, opened 2026-09-10
SED-4923 Capture various execution timestamps as variables

cl-exense:


step-plans/step-plans-core/src/main/java/step/core/execution/ExecutionEngineRunner.java
```diff
@@ -247,8 +257,10 @@ private ReportNode execute(Plan plan, ReportNode rootReportNode) throws Provisio
                 // Do not update the status if the execution was aborted
                 updateStatus(ExecutionStatus.RUNNING);
             }
+            ExecutionTimings.recordTimestamp(executionContext, ExecutionTimings.TimestampVar.STEP_EXEC_TIMESTAMP_START);
             return artefactHandlerManager.execute(root, rootReportNode, ParentSource.MAIN);
         } finally {
+            ExecutionTimings.recordTimestamp(executionContext, ExecutionTimings.TimestampVar.STEP_EXEC_TIMESTAMP_END);
```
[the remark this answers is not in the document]
cl-exense: No. This method does not throw "unexpected exceptions", otherwise we can clutter the entire codebase with try-catch constructs for "unexpected exceptions".

step-plans/step-plans-core/src/main/java/step/core/execution/model/ExecutionTimings.java
```diff
@@ -0,0 +1,60 @@
...
+        STEP_EXEC_TIMESTAMP_EXPORT,
+    }
+
+    public static void recordTimestamp(ExecutionContext context, TimestampVar timing) {
+        // Won't happen with current code, but for good measure:
+        Objects.requireNonNull(context, "context must not be null");
+        Objects.requireNonNull(timing, "timing must not be null");
+
+        // IS0-8601 format, e.g. 2026-09-08T13:09:54.740868590+02:00
+        String isoLocal = OffsetDateTime.now().toString();
```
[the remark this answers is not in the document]
cl-exense: No. We explicitly WANT the time in the configured local timezone, as it's both more readable for humans and conveys additional information. If an external system is incapable of correctly handling timezone offsets, then it's that system's problem, not ours.

step-plans/step-plans-core/src/main/java/step/core/execution/model/ExecutionTimings.java
```diff
@@ -0,0 +1,60 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import step.core.execution.ExecutionContext;
+
+import java.time.OffsetDateTime;
+import java.util.Objects;
+
+/**
+ * This class was introduced as part of the JBS-251 feature request,
```
david-stephan: Minor, but we should reference internal development ticket rather than support ticket

step-plans/step-plans-core/src/main/java/step/core/execution/model/ExecutionTimings.java
```diff
@@ -0,0 +1,60 @@
...
+
+    public static void recordTimestamp(ExecutionContext context, TimestampVar timing) {
+        // Won't happen with current code, but for good measure:
+        Objects.requireNonNull(context, "context must not be null");
+        Objects.requireNonNull(timing, "timing must not be null");
+
+        // IS0-8601 format, e.g. 2026-09-08T13:09:54.740868590+02:00
+        String isoLocal = OffsetDateTime.now().toString();
+        // Variables are saved to the root node
+        context.getVariablesManager().putVariable(context.getReport(), timing.name(), isoLocal);
```
david-stephan, line 55: Why adding them as variable to the plan and not directly to the ExecutionContext ? Except if we really need them available as variable during execution so that they can be used by plan's controls, I would only add it to the ExecutionContext.

david-stephan (changes requested):

step-plans/step-plans-core/src/main/java/step/core/execution/model/ExecutionTimings.java
cl-exense: Because that's where the Jira/Xray integration looks for data: https://step.dev/knowledgebase/plugins/jira-xray/#issuecreationfields . It's using variables because that's how users can set the content themselves (which was obviously the original intention, we're just piggy-backing on top of the already existing functionality). 

Introducing some other special lookup method just for this information seems counterproductive (and besides, we're already using the variables manager to pass around quite a few "internal" metadata..., plus as mentioned this won't harm but might actually be useful information in other contexts).

david-stephan: OK, with the clarifications

david-stephan (approved):

PR #709, opened 2026-09-11
SED-4811 Implement a dedicated UI to execute and monitor agentic test generation

jeromecomte:


step-core/src/main/java/step/automation/packages/AutomationPackageArchive.java
```diff
@@ -30,7 +30,7 @@
 public abstract class AutomationPackageArchive implements Closeable {
 
     private static final Logger log = LoggerFactory.getLogger(AutomationPackageArchive.class);
-    public static final List<String> METADATA_FILES = List.of("automation-package.yml", "automation-package.yaml");
+    public static final List<String> METADATA_FILES = List.of("automation-package.yaml", "automation-package.yml");
```
david-stephan, line 33: I'm not sure whether the change of the order is strictly required. I assume that if an AP has the 2 files ("automation-package.yaml" AND "automation-package.yml") in his packages (unlikely) the one selected will change.

step-ide/src/main/java/step/ide/LocalIDEState.java
```diff
@@ -229,9 +230,22 @@ public void setExecutorDelegateFactory(IDEExecutorDelegateFactory executorDelega
 
     @Override
     public String divertExecution(ExecutionParameters executionParams) {
-        Objects.requireNonNull(currentAutomationPackageDirectory, "currentAutomationPackageDirectory not set; select an AP first");
-        logger.info("Launching diverted execution for parameters: {}", Failable.call(() -> new ObjectMapper().writeValueAsString(executionParams)));
-        IDEExecutorDelegate executorDelegate = executorDelegateFactory.createIDEExecutorDelegate(currentAutomationPackageDirectory.toFile(), executionParams);
+        Path apDir = requireCurrentAutomationPackageDirectory();
+        String description = executionParams.getDescription();
+        List<String> includedPlanNames = (description == null || description.isBlank()) ? List.of() : List.of(description);
```
david-stephan, line 235: I now this is only refactoring and not introduced by this PR, but don't we have a more deterministic way to now which plan has to be executed rather than to rely on the description ?

david-stephan (approved):
@jeromecomte PR OK, and approved, but 2 minor comments to be checked.

PR #712, opened 2026-09-14
SED-4855 decommission-the-keyword-packages

david-stephan:


step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/embedded/EmbeddedAutomationPackagePlugin.java
```diff
@@ -0,0 +1,61 @@
...
+
+        String packageFolder = context.getConfiguration().getProperty(CONFIGURATION_EMBEDDED_FOLDER);
+        if (packageFolder == null) {
+            logger.debug("No embedded automation package folder configured under {}.", CONFIGURATION_EMBEDDED_FOLDER);
+        } else {
+            new EmbeddedAutomationPackageImporter(
+                    context.require(AutomationPackageManager.class),
+                    context.get(ObjectHookRegistry.class),
+                    context.require(AttributeResolverRegistry.class)
+            ).importEmbeddedAutomationPackages(packageFolder);
```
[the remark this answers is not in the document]
david-stephan: agreed and updated

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageMigrationPlugin.java
```diff
@@ -0,0 +1,97 @@
...
+        new KeywordPackageMigrationExecutor(
+                context.getCollectionFactory().getCollection(
+                        KeywordPackageMigrationTask.STAGING_COLLECTION, StagedKeywordPackage.class),
+                context.require(FunctionAccessor.class),
+                context.require(AutomationPackageAccessor.class),
+                context.require(AutomationPackageManager.class),
+                context.getResourceManager(),
+                context.get(ObjectHookRegistry.class),
+                mode
+        ).run();
```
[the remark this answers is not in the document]
david-stephan: agreed and updated

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageMigrationExecutor.java
```diff
@@ -0,0 +1,299 @@
...
+    private AutomationPackageFileSource sourceFor(String location, String resourceType) throws Exception {
+        if (location.startsWith(FileResolver.RESOURCE_PREFIX)) {
+            String resourceId = FileResolver.resolveResourceId(location);
+            retypeResource(resourceId, resourceType);
+            return AutomationPackageFileSource.withResourceId(resourceId);
+        }
+        File file = new File(location);
+        try (InputStream content = new FileInputStream(file)) {
+            return AutomationPackageFileSource.withInputStream(content, file.getName());
+        }
```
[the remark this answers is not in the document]
david-stephan: The inputStream needs to be read later during deployment, ths is correct. However it won't be closed by the AP manager. Refactored the colde to open/close the stream in the deploy function (wrapping the full deployment)

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/embedded/EmbeddedAutomationPackageDescriptor.java
```diff
@@ -0,0 +1,38 @@
+/*******************************************************************************
+ * Copyright (C) 2020, exense GmbH
```
cl-exense, line 2: We keep adding new files with completely outdated dates and redundant copyright notices. I've mentioned multiple times that this is counterproductive, but I basically already gave up... ;-)

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/embedded/EmbeddedAutomationPackageImporter.java
```diff
@@ -0,0 +1,162 @@
...
+
+    private final AutomationPackageManager automationPackageManager;
+    private final ObjectHookRegistry objectHookRegistry;
+    private final AttributeResolverRegistry attributeResolverRegistry;
+    private final ObjectMapper objectMapper = new ObjectMapper();
+
+    public EmbeddedAutomationPackageImporter(AutomationPackageManager automationPackageManager,
+                                             ObjectHookRegistry objectHookRegistry,
+                                             AttributeResolverRegistry attributeResolverRegistry) {
+        this.automationPackageManager = automationPackageManager;
```
cl-exense: I know I'm being a PITA, but I assume all of these should be Objects.requireNonNull.

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/embedded/EmbeddedAutomationPackageImporter.java
```diff
@@ -0,0 +1,162 @@
...
+        this.objectHookRegistry = objectHookRegistry;
+        this.attributeResolverRegistry = attributeResolverRegistry;
+    }
+
+    /**
+     * @return the ids of the deployed automation packages
+     */
+    public List<String> importEmbeddedAutomationPackages(String packageFolder) {
```
cl-exense, line 82: The argument should be checked to not be null. In addition, this will cause RuntimeExceptions (supposedly not very readable ones) if the given path (or one of the "local"/"remote" subpaths) exists, but is not a directory but a file.

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageClassifier.java
```diff
@@ -0,0 +1,91 @@
...
+     * absolute path and will now be managed by the embedded Automation Packages, so they are deleted rather than migrated in every mode.
+     */
+    static final String EMBEDDED_PACKAGE_CUSTOM_FIELD = "embeddedPackage";
+
+    private final Predicate<String> fileReadable;
+
+    public KeywordPackageClassifier() {
+        this(path -> {
+            File file = new File(path);
+            return file.exists() && file.canRead();
```
cl-exense: You'll probably also want to check that it's actually a file, not a directory.

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageBucket.java
```diff
@@ -0,0 +1,46 @@
...
+ * along with STEP.  If not, see <http://www.gnu.org/licenses/>.
+ ******************************************************************************/
+package step.automation.packages.migration;
+
+/**
+ * What has to become of a keyword package. Only the distinctions the migration acts on are modelled:
+ * whether the archive sits in a resource or on the filesystem is settled at deployment time, from the
+ * location itself.
+ */
+public enum KeywordPackageBucket {
```
cl-exense: "Bucket" is a slightly misleading term, as a bucket usually is a container that holds related items, not just an "attribute"/discriminator. I suggest "kind" instead ("type" or "class" would also do, but already have specific java language meanings).

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageClassifier.java
```diff
@@ -0,0 +1,91 @@
...
+        });
+    }
+
+    /**
+     * @param fileReadable decides whether a filesystem location still resolves
+     */
+    public KeywordPackageClassifier(Predicate<String> fileReadable) {
```
cl-exense: Suggest to make this package-private. It would be highly unusual to have "real" code that provides its own predicate implementation, and the unit test can still access the relevant constructur.

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageBucket.java
```diff
@@ -0,0 +1,46 @@
...
+    /** Everything it needs is reachable, so it is replaced by an automation package. */
+    MIGRATABLE,
+
+    /**
+     * Something the package needs is not there: its archive, or the libraries it declares. Libraries
+     * count because the keywords were built against them, so a package converted without them would be
+     * one that cannot run. Nothing can execute these keywords today either way, so the package is
+     * logged with its location and removed without a replacement.
+     */
+    ARCHIVE_MISSING
```
cl-exense: If "something" is missing then INCOMPLETE may be a better name than the very specific ARCHIVE_MISSING name.

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageClassifier.java
```diff
@@ -0,0 +1,91 @@
...
+
+        return KeywordPackageBucket.MIGRATABLE;
+    }
+
+    /**
+     * A {@code resource:} reference is taken at face value here. Whether the resource still exists is a
+     * database question, and this classifier deliberately has none, so that is checked when the package
+     * is converted — where a missing resource leads to the same outcome as a missing file.
+     */
+    private boolean isUnavailable(String location) {
```
cl-exense, line 84: Nitpick: this code is hard to understand because everything (both the method name as such, and the checks it performs) is negated.

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageMigrationExecutor.java
```diff
@@ -0,0 +1,310 @@
...
+    private final KeywordPackageClassifier classifier;
+
+    public KeywordPackageMigrationExecutor(Collection<StagedKeywordPackage> staging,
+                                           FunctionAccessor functionAccessor,
+                                           AutomationPackageAccessor automationPackageAccessor,
+                                           AutomationPackageManager automationPackageManager,
+                                           ResourceManager resourceManager,
+                                           ObjectHookRegistry objectHookRegistry,
+                                           KeywordPackageMigrationMode mode) {
+        this.staging = staging;
```
cl-exense: Again, this is a public method, so it should verify that arguments which are required to be non-null actually are not null. I know it's tedious, but that's the Java reality.

(for fun: https://qconlondon.com/london-2009/qconlondon.com/london-2009/presentation/Null%2BReferences_%2BThe%2BBillion%2BDollar%2BMistake.html )

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageMigrationExecutor.java
```diff
@@ -0,0 +1,310 @@
...
+        this.resourceManager = resourceManager;
+        this.objectHookRegistry = objectHookRegistry;
+        this.mode = mode;
+        this.classifier = new KeywordPackageClassifier();
+    }
+
+    public void run() {
+        List<StagedKeywordPackage> staged = staging.find(Filters.empty(), null, null, null, 0).toList();
+        if (staged.isEmpty()) {
+            staging.drop();
```
cl-exense, line 99: Add a comment that this is because non-existent collections are silently recreated on access by the framework.

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageMigrationExecutor.java
```diff
@@ -0,0 +1,310 @@
...
+            try {
+                if (process(stagedPackage, bucket)) {
+                    deployed++;
+                }
+            } catch (Exception e) {
+                failures.add(stagedPackage.describe());
+                logger.error("Unable to migrate the keyword package {}. It has been removed without a "
+                        + "replacement.", stagedPackage.describe(), e);
+            } finally {
+                // A failed package is not retried, so its record goes either way.
```
cl-exense: "its record goes either way" is doubly misunderstandable. Better phrasing is "its record is deleted regardless/in any case/anyway".

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageMigrationExecutor.java
```diff
@@ -0,0 +1,310 @@
...
+        } else if (mode == KeywordPackageMigrationMode.DETACH) {
+            detachKeywords(staged);
+        } else if (KeywordPackageBucket.ARCHIVE_MISSING.equals(bucket)) {
+            // The resources are kept: the missing one may be only the archive or only the libraries,
+            // and the other can still be a valid resource shared with another package.
+            logger.warn("The keyword package {} points at an archive or libraries that cannot be read, "
+                    + "its keywords cannot be executed. The package and its keywords are removed "
+                    + "without a replacement, its Step resources are left in place.", staged.describe());
+            deleteKeywords(staged);
+        } else {
```
cl-exense: Add a comment stating which cases (mode/KW kind) that represents. It's unclear from this code alone.

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageMigrationExecutor.java
```diff
@@ -0,0 +1,310 @@
...
+        }
+        if (!candidate.equals(baseName)) {
+            logger.warn("An automation package named '{}' already exists, the keyword package {} is "
+                    + "migrated as '{}'.", baseName, staged.describe(), candidate);
+        }
+        return candidate;
+    }
+
+    private boolean isNameTaken(String name, ObjectPredicate objectPredicate) {
+        return java.util.stream.StreamSupport
```
cl-exense: Unnecessarily fully qualified. Use import instead.

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageMigrationExecutor.java
```diff
@@ -0,0 +1,310 @@
...
+        List<Function> keywords = keywordsOf(staged);
+        for (Function keyword : keywords) {
+            keyword.getCustomFields().remove(FUNCTION_PACKAGE_ID_CUSTOM_FIELD);
+            functionAccessor.save(keyword);
+        }
+        logger.info("Detached {} keyword(s) from the keyword package {}.", keywords.size(), staged.describe());
+    }
+
+    /**
+     * Removed through the accessor and not {@code FunctionManager.deleteFunction}, which delegates to
```
cl-exense: I don't understand this comment, please rephrase :)

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageMigrationExecutor.java
```diff
@@ -0,0 +1,310 @@
...
+    private List<Function> keywordsOf(StagedKeywordPackage staged) {
+        return functionAccessor.findManyByCriteria(
+                        Map.of("customFields." + FUNCTION_PACKAGE_ID_CUSTOM_FIELD, staged.getId().toString()))
+                .collect(Collectors.toList());
+    }
+
+    /**
+     * Does nothing for a package located by a filesystem path, which owns no resource.
+     */
+    private void deleteResource(String location) {
```
cl-exense: given the behavior, deleteIfIsResource seems to be a better name.

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageMigrationMode.java
```diff
@@ -0,0 +1,93 @@
...
+     *                                  the manual reconfiguration that mode exists to protect
+     */
+    public static KeywordPackageMigrationMode parse(String propertyValue) {
+        if (propertyValue == null || propertyValue.isBlank()) {
+            return MIGRATE;
+        }
+        String trimmed = propertyValue.trim();
+        try {
+            // Locale.ROOT rather than the default locale: "migrate" contains an 'i', which a Turkish
+            // locale would upper-case to 'İ' and fail to match.
```
cl-exense, line 82: :-D The joys of locales. Nice example of why it's important even for plain ASCII...

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageMigrationPlugin.java
```diff
@@ -0,0 +1,97 @@
...
+ * <p>The 2nd phase is executed in the afterInitializeData of the controller plugin using the Resource and
+ * Automation Package managers
+ *</p>
+ * <p>Because it drains the staging collection on every startup rather than once, an interruption
+ * between the two phases is picked up on the next boot even though the migration task itself will not
+ * run again.
+ * </p>
+ */
+@Plugin(dependencies = {MigrationManagerPlugin.class, AutomationPackagePlugin.class})
+public class KeywordPackageMigrationPlugin extends AbstractControllerPlugin {
```
cl-exense, line 58: If I understand correctly, this is a plugin that will basically run perpetually on every start, even though the actual migration should really be a one-time task. Why is it required to be implemented this way?

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageMigrationPlugin.java
```diff
@@ -0,0 +1,97 @@
...
+     * A rejected value has to stop the startup rather than fall back to the default. Only a
+     * {@link PluginCriticalException} propagates out of a plugin lifecycle method — anything else is
+     * logged and swallowed — so a mistyped {@code detach} would otherwise boot straight into the mode
+     * that rebuilds keywords and discards their manual reconfiguration.
+     */
+    private KeywordPackageMigrationMode resolveMode(GlobalContext context) {
+        try {
+            KeywordPackageMigrationMode mode = KeywordPackageMigrationMode.parse(
+                    context.getConfiguration().getProperty(KeywordPackageMigrationMode.PROPERTY_KEY));
+            logger.info("Keyword package migration mode: {}", mode.getPropertyValue());
```
cl-exense: I understand that this is useful information, but this is really just noise that will be logged on every startup, even if it's only relevant a single time. Relates to the above comment about why the plugin is necessary at all :-)

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageMigrationTask.java
```diff
@@ -0,0 +1,79 @@
...
+            // automatically creates the table for some factory, so we immediately drop it again
+            logger.info("No keyword package found, dropping the collection.");
+            keywordPackages.drop();
+            return;
+        }
+
+        // Rename the collection for the actual migration/cleanup
+        keywordPackages.rename(STAGING_COLLECTION);
+
+        logger.info("Moved {} keyword package(s) aside for the migration to automation packages.", count);
```
cl-exense: "aside" is too colloquial and not informative. Suggestion: "Collection {} was renamed to {} for upcoming migration of {} keyword packages to automation packages", ...

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/StagedKeywordPackage.java
```diff
@@ -0,0 +1,102 @@
...
+
+import step.core.accessors.AbstractOrganizableObject;
+import step.core.objectenricher.EnricheableObject;
+
+import java.util.Map;
+
+/**
+ * A keyword package waiting to be replaced, as moved aside by {@link KeywordPackageMigrationTask}.
+ * <p>
+ * This is a verbatim copy of the {@code functionPackage} document: the migration task moves the
```
cl-exense: I don't understand the first sentence. What `functionPackage` document?
...
OK, I think I got it. But "verbatim" is the wrong word, as the class definition is substantially different from the previous one, and it's not one document. Suggest something like:

"This class is structurally compatible with the data previously present in the `functionPackage` collection that is about to be migrated; the migration task moves..."

step-controller/step-controller-server/src/main/java/step/core/deployment/ObjectHookControllerPlugin.java
```diff
@@ -42,6 +43,10 @@ public void serverStart(GlobalContext context) throws Exception {
         ObjectPredicateFactory objectPredicateFactory = new ObjectPredicateFactory(objectHookRegistry);
         context.put(ObjectPredicateFactory.class, objectPredicateFactory);
 
+        // Registered here so that every plugin contributing or consuming a resolver is ordered after
+        // this one, as they already are for the hook registry.
```
cl-exense: What? :-D This comment is confusing me more than it helps ;-)
... edit: I suggest to rephrase to something like: "Instantiating here, as plugins using the AttributesResolverRegistry already depend on this plugin anyway because of the hook registry."

step-core-model/src/main/java/step/core/objectenricher/AttributeResolverRegistry.java
```diff
@@ -0,0 +1,66 @@
...
+ * Resolves symbolic attribute values written as {@code @<value>} into the value to be persisted.
+ * <p>
+ * Used for attributes an administrator writes by hand, outside of any project context: the embedded
+ * automation package sidecars name their project as {@code "@Common"} while the persisted attribute
+ * is the project id. Enterprise registers the resolver for {@code project}; without it the values
+ * are left untouched.
+ */
+public class AttributeResolverRegistry {
+
+    private static final String RESOLVABLE_PREFIX = "@";
```
cl-exense, line 36: This is really a micro-optimization, but we can use a character here.

step-core-model/src/main/java/step/core/objectenricher/AttributeResolverRegistry.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * @return the resolved value, or the value unchanged when it carries no prefix or when no
+     * resolver is registered for the key
+     */
+    public String resolve(String key, String value) {
+        Function<String, String> resolver = resolvers.get(key);
+        String resolved;
+        if (resolver != null && value != null && value.startsWith(RESOLVABLE_PREFIX)) {
+            resolved = resolver.apply(value.substring(RESOLVABLE_PREFIX.length()));
```
cl-exense, line 52: ... and if using a character, we don't have to use the expensive string operations and calculate length() all over, because we know it's 1 :)

step-core-model/src/main/java/step/core/objectenricher/AttributeResolverRegistry.java
```diff
@@ -0,0 +1,66 @@
...
+        } else {
+            resolved = value;
+        }
+        return resolved;
+    }
+
+    public Map<String, String> resolveAll(Map<String, String> attributes) {
+        Map<String, String> resolved = new LinkedHashMap<>();
```
cl-exense, line 60: Be aware that LinkedHashMap preserves the order values were inserted -- which is a good thing, but it also depends on the traversal order of the map that's being passed in.

step-core/src/main/java/step/resources/ResourceManager.java
```diff
@@ -162,6 +162,18 @@ Resource saveResourceContent(Resource resource, InputStream resourceStream, Stri
      */
     Resource saveResource(Resource resource) throws IOException;
 
+    /**
+     * Changes the type of an existing resource, relocating its content accordingly: the content is
+     * stored per type, so updating the type through {@link #saveResource(Resource)} alone leaves the
```
cl-exense, line 167: This comment needs a little more explanation IMO.

"... alone would leave the content where it can no longer be found. Thus, if the type of an existing resource needs to be changed while keeping its id, use this method instead of `saveResource`."

step-core/src/main/java/step/resources/ResourceManagerImpl.java
```diff
@@ -417,6 +417,36 @@ public Resource saveResource(Resource resource) throws IOException {
...
+        if (!resourceTypes.containsKey(resourceType)) {
+            throw new RuntimeException("Unknown resource type " + resourceType);
+        }
+        Resource resource = getResource(resourceId);
+        if (!resourceType.equals(resource.getResourceType())) {
+            File currentContainer = getResourceContainer(resource);
+            String previousType = resource.getResourceType();
+            resource.setResourceType(resourceType);
+            File newContainer = getResourceContainer(resource);
+            boolean moved = currentContainer.exists();
```
cl-exense: minor: moved sounds like the past tense, suggest to use "moving" instead.
Anyway, what if it doesn't exist? Then obviously nothing will be moved, but can this be a problem?

cl-exense (changes requested):

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/embedded/EmbeddedAutomationPackageDescriptor.java
david-stephan: All our OSS source files shall contain our copyright and should be aligned. If it is outdated it should be updated in a bulk, not for individual files and not as part of a feature PR.

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/embedded/EmbeddedAutomationPackageImporter.java
david-stephan: updated

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/embedded/EmbeddedAutomationPackageImporter.java
david-stephan: updated

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageBucket.java
david-stephan: updated

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageBucket.java
david-stephan: updated

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageClassifier.java
david-stephan: updated

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageClassifier.java
david-stephan: updated

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageClassifier.java
david-stephan: It was implemented as "isAvailable" but all usages were then negated, which I don't really like either (and intellij complains about it)...

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageMigrationExecutor.java
david-stephan: updated

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageMigrationExecutor.java
david-stephan: updated

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageMigrationExecutor.java
david-stephan: updated

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageMigrationExecutor.java
david-stephan: updated

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageMigrationExecutor.java
david-stephan: updated with a if block

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageMigrationExecutor.java
david-stephan: updated

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageMigrationExecutor.java
david-stephan: updated

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageMigrationTask.java
david-stephan: updated

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/StagedKeywordPackage.java
david-stephan: updated

step-controller/step-controller-server/src/main/java/step/core/deployment/ObjectHookControllerPlugin.java
david-stephan: updated, with something a bit different, hopefully clear enough

step-core-model/src/main/java/step/core/objectenricher/AttributeResolverRegistry.java
david-stephan: not changed, prefer to keep "startsWith" rather than chartAt which would also require handling of empty strings.

step-core/src/main/java/step/resources/ResourceManager.java
david-stephan: If think your suggestion don't had more details but emphasize that saveRessource should not be used to modify the resource's type, made an update in that sense

step-core/src/main/java/step/resources/ResourceManagerImpl.java
david-stephan: udpated

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageMigrationPlugin.java
david-stephan: was initially done to support interruption/crash but is not required, refactored.

step-automation-packages/step-automation-packages-controller/src/main/java/step/automation/packages/migration/KeywordPackageMigrationPlugin.java
david-stephan: refactored

cl-exense (dismissed):

PR #713, opened 2026-09-16
SED-4941 Create a setting to disable legacy report view

neogucky:


david-stephan (approved):
@neogucky PR approved. The Frontend is the only consumer of these configuration items, si I don't mind the gemini feedback as long as it doesn't break the FE.

PR #722, opened 2026-09-18
SED-4811 Implement a dedicated UI to execute and monitor agentic test generation 2

jeromecomte:


step-core-model/src/main/java/step/core/yaml/YamlMetadata.java
```diff
@@ -0,0 +1,92 @@
...
+     * @return the metadata of the source or null if there is none
+     */
+    public static Map<String, Object> extractFrom(AbstractIdentifiableObject source) {
+        return source.getMetadata();
+    }
+
+    /**
+     * @throws IllegalArgumentException if one of the keys (at any depth) is invalid
+     */
+    public static void validate(Map<String, Object> metadata) {
```
david-stephan: this seems to be only called by this class and should be private

step-core-model/src/main/java/step/core/yaml/YamlMetadata.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Pattern;
+
+/**
+ * The free-form {@code metadata} that can be declared on the automation package itself and on each of its entities
+ * (keywords, plans, parameters, schedules...). The YAML content is kept as it is, as a map, and persisted in the
+ * {@link AbstractIdentifiableObject#getMetadata()} of the business object.
+ */
+public class YamlMetadata {
```
david-stephan, line 32: Something feels not right with this helper class. We already use custom (de)serilaizer for example in YamlPlan with:
```
    @JsonSerialize(using = AgentProvisioningConfigurationSerializer.class)
    @JsonDeserialize(using = AgentProvisioningConfigurationDeserializer.class)
    private AgentProvisioningConfiguration agents;
```
Why cannot we just use a custom deseralizer?

step-core-model/src/main/java/step/core/yaml/YamlMetadata.java
```diff
@@ -0,0 +1,92 @@
...
+     */
+    public static void applyTo(AbstractIdentifiableObject target, Map<String, Object> metadata) {
+        if (metadata != null && !metadata.isEmpty()) {
+            validate(metadata);
+            target.setMetadata(metadata);
+        }
+    }
+
+    /**
+     * @return the metadata of the source or null if there is none
```
david-stephan, line 62: the comment tends to say there is some logic apply, it just return the source value

david-stephan (commented):
Asking for clarifications and potential changes. IMO setting a custom deserializer on the few YAML model classes directly would simplify a lot the code / PR.

step-core-model/src/main/java/step/core/yaml/YamlMetadata.java
jeromecomte: @david-stephan The code of this class is not about serialization from YAML to YAML Pojo. It is about YAML Pojo to "DB" Pojo. I couldn't find any central place that would allow to define this mapping. The class doesn't do more than just validating and mapping the field <YAML Pojo>.metadata to AbstractIdentifiableObject.metadata but I wouldn't now how and where do to this instead

jeromecomte: As discussed, we should simplify the whole mapping process in the future

david-stephan (approved):

